PR #12, opened 2015-01-08
Add in call backs for drag/resize calls

browniefed:
Thoughts on adding callbacks for this stuff?
I just basically passed through most information since I wasn't sure exactly what people were going to want to utilize.

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,6 +183,18 @@ listenToWindowResize: React.PropTypes.bool,
 // Callback so you can save the layout.
 // Calls back with (currentLayout, allLayouts). allLayouts are keyed by breakpoint.
 onLayoutChange: React.PropTypes.func,
+// Calls when drag starts. Callback is of the signature (newLayout, oldLayout, layoutItem, placeholder, {e, element, position}.
+onDragStart: React.PropTypes.func,
+// Calls on each drag movement. Callback is of the signature (newLayout, oldLayout, layoutItem, placeholder, {e, element, position}.
+onDrag: React.PropTypes.func,
+// Calls when drag is complete. Callback is of the signature (newLayout, oldLayout, layoutItem, placeholder, {e, element, position}.
+onDragStop: React.PropTypes.func,
+//Calls when resize starts. Callback is of the signature (newLayout, oldLayout, layoutItem, placeholder, {e, element, size}.
+onResizeStart: React.PropTypes.func,
+// Calls when resize movement happens. Callback is of the signature (newLayout, oldLayout, layoutItem, placeholder, {e, element, size}.
+onResize: React.PropTypes.func,
+// Calls when resize is complete. Callback is of the signature (newLayout, oldLayout, layoutItem, placeholder, {e, element, size}.
+onResizeStop: React.PropTypes.func
 ```
 
 #### Responsive Grid Layout Props
--- a/build/GridItem.js
+++ b/build/GridItem.js
@@ -231,7 +231,7 @@ var GridItem = React.createClass({
       // Cap x at numCols
       x = Math.min(x, me.props.cols - me.props.w);
 
-      me.props[handlerName](me.props.i, x, y);
+      me.props[handlerName](me.props.i, x, y, { e: e, element: element, position: position });
     };
   },
 
@@ -268,7 +268,7 @@ var GridItem = React.createClass({
 
       me.setState({ resizing: handlerName === "onResizeStop" ? null : size });
 
-      me.props[handlerName](me.props.i, w, h);
+      me.props[handlerName](me.props.i, w, h, { e: e, element: element, size: size });
     };
   },
 
--- a/lib/ReactGridLayout.jsx
+++ b/lib/ReactGridLayout.jsx
@@ -61,7 +61,18 @@ var ReactGridLayout = React.createClass({
     // Callback so you can save the layout.
     // Calls back with (currentLayout, allLayouts). allLayouts are keyed by breakpoint.
     onLayoutChange: React.PropTypes.func,
-
+    // Calls when drag starts
+    onDragStart: React.PropTypes.func,
+    // Calls on each drag movement
+    onDrag: React.PropTypes.func,
+    // Calls when drag is complete
+    onDragStop: React.PropTypes.func,
+    //Calls when resize starts
+    onResizeStart: React.PropTypes.func,
+    // Calls when resize movement happens
+    onResize: React.PropTypes.func,
+    // Calls when resize is complete
+    onResizeStop: React.PropTypes.func,
 
     //
     // Other validations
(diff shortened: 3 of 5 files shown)


build/GridItem.js
```diff
@@ -70,7 +70,7 @@ var GridItem = React.createClass({
     handle: React.PropTypes.string
   },
 
-  getDefaultProps: function () {
+  getDefaultProps: function getDefaultProps() {
```
STRML: Which version of 6to5 are you using? I wonder why the build came out this way for you; this is definitely better for stack traces.

lib/ReactGridLayout.jsx
```diff
@@ -57,7 +57,16 @@ var ReactGridLayout = React.createClass({
...
     onLayoutChange: React.PropTypes.func,
-
+    // Calls when drag starts, allows for canceling/restricting drag
+    onDragStart: React.PropTypes.func,
+    // Calls on each drag movement, allows for modifying current drag layout, or preventing layout modificaiton
+    onDrag: React.PropTypes.func,
+    // Cals when drag is complete
+    onDragStop: React.PropTypes.func,
+    // Calls when resize starts, allows to cancel/restrict resize
+    onResize: React.PropTypes.func,
```
STRML: I suppose we should add an onResizeStart to be consistent?

lib/ReactGridLayout.jsx
```diff
@@ -152,11 +166,33 @@ var ReactGridLayout = React.createClass({
...
+   * @param {Number} i Index of the child
+   * @param {Number} x X position of the move
+   * @param {Number} y Y position of the move
+   * @param {Element} element The current dragging DOM element
+   * @param {Object} position Drag information
+   */
+  onDragStart(i, x, y, {e, element, position}) {
+    var layout = this.state.layout;
+    var l = utils.getLayoutItem(layout, i);
+    if (this.props.onDragStart(layout, this.state.layout, l, null, {e, element, position}) === false) {
```
STRML: I like making it cancelable via the callback; any ideas how this is done in other libraries? My instinct is to check if the event has been canceled (e.g. e.preventDefault()) but that doesn't quite make as much sense in React.

STRML: Of course if we did that we'd have to be consistent among all the other callbacks as well. With this code as-is the placeholder would quickly be replaced by the drag handler below.

build/GridItem.js
browniefed: I just did an npm install so whatever the package.json says

lib/ReactGridLayout.jsx
browniefed: I'll add it

lib/ReactGridLayout.jsx
browniefed: Yeah, I think e.preventDefault would be the way to go since I'm not necessarily a fan of returning false, and as of React .12 you must call preventDefault to stop.
It looks like we could do it with mouseDown event since that is what react-draggable is using, however we may need to mod the handleStartDrag to actually do the preventDefault check.

build/GridItem.js
STRML: Gotcha. I had an old version in npm cache. Could you rebase to master so we can squash the diffs?

lib/ReactGridLayout.jsx
STRML: That's fine by me, I'll add that to my fork of draggable.

build/GridItem.js
browniefed: Will do, I'm at work now so I'll handle all these sometime later tonight.

STRML: Cool. Thanks a lot for the contribution, I appreciate it.

browniefed:
Rebased, added onResizeStart, add README documentation.
Only thing this will depend on is react-draggable and react-resizable stopping drag/resize when `preventDefault` gets called.

browniefed:
Hey just curious if you had an idea of when these could get merged?

STRML:
Hey Jason - sorry, chasing down a lot of things on a lot of projects
right now. Should be merged soon, in the meantime I would recommend
keeping this rebased on master and using a github url in your project
until I can finish reviewing.

browniefed:
@STRML no problem, thanks

lib/ReactGridLayout.jsx
```diff
@@ -201,13 +258,20 @@ var ReactGridLayout = React.createClass({
     var activeDrag = {
       w: w, h: h, x: l.x, y: l.y, placeholder: true, i: i
     };
+
+    activeDrag = this.props.onResize(layout, layout, l, activeDrag, {e, element, size}) || activeDrag;
```
STRML: In general I don't really like the assignment here; I don't see any demand for it and it is confusing - on dragStart and resizeStart, returning a value does nothing, on drag, you return the layout, on resize, you return the resizing item, and on dragStop and resizeStop you return the layout again.

Do you have a particular need for that functionality? Otherwise let's just remove the assignment altogether.

README.md
```diff
@@ -181,6 +181,18 @@ listenToWindowResize: React.PropTypes.bool,
 // Callback so you can save the layout.
 // Calls back with (currentLayout, allLayouts). allLayouts are keyed by breakpoint.
 onLayoutChange: React.PropTypes.func,
+// Calls when drag starts, allows for canceling/restricting drag    
+onDragStart: React.PropTypes.func,
+// Calls on each drag movement, allows for modifying current drag layout, or preventing layout modificaiton
+onDrag: React.PropTypes.func,
+// Calls when drag is complete
+onDragStop: React.PropTypes.func,
+//Calls when resize starts, allows for canceling
```
STRML: Could you please fix the spacing here and note the callback signature on each one both here and in the source?

For example: 

"Calls when resize starts. Callback is of the signature `(newLayout, oldLayout, layoutItem, {e, element, size}`."

Modify the callback signature to `{e, element, position}` for drags.

lib/ReactGridLayout.jsx
browniefed: Well onResizeStart/onDragStart you ideally would be canceling with preventDefault so no return necessary.
Ideally if there were a way to set both full layout/active drag layout items that would be ideal.
I'm attempting to implement a lot of restrictions around dragging/resizing  so controlling the full layout/actively dragging/resizing item is what I need.

I realize I can modify the layout and activeDrag by just simple object reference but that's just bad. Besides calling a setState of ReactGridLayout and passing down new props layout during the drag (which would probably be very inefficient and cause some weird UI states temporarily) I'm not sure what else is the solution here.

Basically how would you allow the user to control the positioning of the entire layout/actively dragging item while it's happening?

browniefed:
@STRML hey I just decided to remove the assignment stuff for now, so I've rebased and squashed, fixed the documentation text. I think it should be good to go.

STRML:
Perfect. This is simple, we can worry about the assignment stuff later on.

STRML:
`0.7.1`

PR #13, opened 2015-01-11
Allow each grid item to define whether or not it is draggable/resizable ...

browniefed:
...or not.

### DIFF
--- a/build/ReactGridLayout.js
+++ b/build/ReactGridLayout.js
@@ -284,8 +284,8 @@ var ReactGridLayout = React.createClass({
         onDrag: this.onDrag,
         onResize: this.onResize,
         onResizeStop: this.onResizeStop,
-        isDraggable: l["static"] ? false : this.props.isDraggable,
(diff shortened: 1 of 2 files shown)


STRML:
See [utils#synchronizeLayoutWithChildren()](https://github.com/STRML/react-grid-layout/blob/master/lib/utils.js#L318) - these properties should also be copied over.

To be honest, since that's not a performance-sensitive function, I think it makes more sense to replace that with an `assign` call instead so we don't have to munge this thing over and over.

I'll make that change first to master so you don't have to.

browniefed:
@STRML sounds good, I wasn't seeing them get copied over when it was defined on the global layout array, didn't check whether or not it was working with defining layouts on the children themselves.

STRML:
No reason to close this, it can be merged after I move to Object.assign() on that bit.

browniefed:
@STRML ha my bad, let me know if you need anything else from me or if it doesn't follow your coding style spec.

PR #15, opened 2015-01-16
Use react directly and require addon modules explicitly

kimagure:
Uses the main React module and requires in the specific needed modules separately.

### DIFF
--- a/build/GridItem.js
+++ b/build/GridItem.js
@@ -1,5 +1,6 @@
 "use strict";
-var React = require("react/addons");
+var React = require("react");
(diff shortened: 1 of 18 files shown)


STRML:
Thanks. This should give us a significant file size benefit when webpacked. I'll make sure it doesn't break anything then merge.

STRML:
I've updated the react-draggable fork and react-resizable to also use the `PureRenderMixin` and `cloneWithProps` directly.

Results show subtraction of 14 modules (from 212 to 198 in the example builds) and about 150kB.

Thanks for this.

kimagure:
:+1:

PR #17, opened 2015-01-19
Adds a note about using custom keys

jeffkole:
If you do not use the index for a grid item key, then the layout cannot be found.  You can explicitly set the layout id using an object property named `i`.

I think this would be valuable to add to the documentation.  Not sure if this is the best place for the info though.


STRML:
Thanks. This is a good addition.

PR #22, opened 2015-01-30
Allow passing cancel handle for react-draggable

bryantsai:
Allow passing cancel handle for react-draggable, in addition to '.react-resizable-handle'.


STRML:
Thanks, works for me.

Issue #25, opened 2015-02-02
Typos in example


TrySpace:
`<div key={1} _grid={{x: 0, y: 0, w: 1: h: 2}}>1</div>
          <div key={2} _grid={{x: 1, y: 0, w: 1: h: 2}}>2</div>
          <div key={3} _grid={{x: 2, y: 0, w: 1: h: 2}}>3</div>`

3 of those semicolons should be commas

STRML:
I don't follow this report at all - what are you saying?

TrySpace:
The object literals, the're not valid json or javascript.

browniefed:
@TrySpace I don't see a single semicolon. That all looks like valid React and JavaScript to me.
A lot of these issues would probably be better as pull requests since that is much clearer on what you're saying is wrong.

TrySpace:
Sorry, I meant colons

PR #28, opened 2015-02-04
Add commas to example to make valid JavaScript

browniefed:
Fix for #25


STRML:
Thanks. That makes more sense now, I was looking straight at it and couldn't see it.

PR #34, opened 2015-02-20
Add new prop to disable vertical compact

TeaBough:
I've added the ability to disable vertical compact.


browniefed:
@TeaBough you should also document this new prop in the Grid Layout Props in the README

TeaBough:
Oups, I forgot. It's done.

STRML:
This is good, but is there a reason why the build files have been renamed?

TeaBough:
It's because the rename command is not working the same way on Archlinux (the one in the Makefile). So I've changed it with a combo find/mv. I think that doing the rename this way is more portable.

STRML:
Thanks for the contribution. Could you please rebase this on master? I've already incorporated your `rename` change and the `build/` directory to clean up diffs. I saw this [great article](http://mammal.io/articles/using-es6-today/#publishing-to-npm) about publishing ES6 projects so I've made some changes.

benmosher:
Should this be closed, on account of being merged (in spirit) via #48?

TeaBough:
Yes

PR #48, opened 2015-04-29
Add property to allow for disabling of vertical compacting

NSinopoli:
I've fixed up #34 for a clean merge.


STRML:
Thanks. I've been thinking that I should simplify the args passed to the utils functions because they're getting a little unwieldy now. I'd like to just pass props and state so we can add properties like this without so many changes.

Anyway that's no reason to delay this. Thanks for the contribution.

PR #54, opened 2015-05-06
updating comments and removing extraneous comma

phchang:
Sorry for the pedantic updates, but came across these as I was going through the code.


STRML:
Thanks.

Issue #53, opened 2015-05-05
Using it with bootstrap, the layout get messy


mamcx:
I try to setup this on top of a bootstrap 3 site, copy the css + sample css, and the grid render weirdly and the draggable handler get rendered out of place.

Any hint in how integrate it with bootstrap?

NSinopoli:
Use the CSS found here:

https://github.com/STRML/react-resizable/blob/master/css/styles.css

benmosher:
I'm confused about why this is closed, but the PR is open and unmerged. This is still an issue in the published version (0.8.5).

Using the `react-resizable` CSS does seem to solve the problem, but it seems like it should either be part of the `react-grid-layout` CSS proper, or `react-resizable` could be a peer dependency maybe?

That said: great module! :+1: not trying to be a downer, but I am confused why this is closed when I just ran into this as well.

STRML:
This is closed because the styles are in `master`. 

This module badly needs an update, I've been way too distracted with my day job / other projects. See https://github.com/mzabriskie/react-draggable/tree/coreRefactor, which once completed will become the next version of `<Draggable>` (which affects `<Resizable>` and this module).

PR #69, opened 2015-06-03
Update styles

NSinopoli:
This should patch #53.


STRML:
These styles are in `master`.

PR #70, opened 2015-06-03
Fix oldItem arguments in resize/drag callbacks

### LINKED ISSUE #52: onResizeStop arguments not correct
For one, the signature for all the event handlers is undocumented.
Looking at the code for onResizeStop, I believe it should be calling my handler with 
0. the current grid layout
1. the previous layout of the box being resized
2. the new layout of the box being resized
3. null
4. the event
However,` oldL = utils.clone(l)`, so arguments 1 and 2 are always identical.

NSinopoli:
This fixes #52.

### DIFF
--- a/lib/ReactGridLayout.jsx
+++ b/lib/ReactGridLayout.jsx
@@ -125,7 +125,9 @@ var ReactGridLayout = React.createClass({
       activeDrag: null,
       isMounted: false,
       layout: utils.synchronizeLayoutWithChildren(this.props.layout, this.props.children, this.props.cols, this.props.verticalCompact),
-      width: this.props.initialWidth
+      width: this.props.initialWidth,
+      oldDragItem: null,
+      oldResizeItem: null
     };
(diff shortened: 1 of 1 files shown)


STRML:
Nice one.

Working on the IE11 compat stuff right now. Getting some weird results on resize. Will merge this after.

NSinopoli:
Thanks!

As an aside, how do I build the dist files locally? Running `npm run build` leaves me with a build that doesn't have the `classnames` copyright information in here:

https://github.com/STRML/react-grid-layout/blob/master/dist/react-grid-layout.min.js

STRML:
That probably has something to do with your webpack version. There's a
mistake in build.sh - it's not invoking the local webpack, but the
system webpack.

NSinopoli:
Ah, I hadn't run `npm install` prior to running `./build.sh`, so it wasn't pulling in `classnames`. Thanks so much!

Issue #104, opened 2015-09-22
horizontal units as percentages


benmosher:
Would you be open to a PR that changes all the width calculations from pixel values to percentages?

Also: any obvious gotchas to a percentage-based column width system? I think it might remove the dependency on the window `resize` event and would empower the browser to animate width changes as necessary.

I am not sure what sort of server-side rendering you're trying to preserve, but assuming the end client is still a browser, I don't see any issues there. 

AFAICT the `transform` styles accept mixed units (i.e. `translate(5%, 14px)`) so I don't think there is any gotcha there, either.

I may need to fork and provide this functionality anyway. Just curious if you'd consider merging it back in and/or you've tried it and it fails for [x reason].

Thanks! 😄

STRML:
I would be 100% open to this as it also makes server-side rendering much more predictable. 

The issue I've found with it is that percentages in `translate()` are relative to the element itself (!), not the viewport, so some fancy math would be necessary in order to ensure that the offset is correct.

See my notes here: https://github.com/STRML/react-grid-layout/blob/master/lib/GridItem.jsx#L174

STRML:
`<Draggable>` would need to be adapted to work with percentages as well. This would require knowledge of the parent's dimensions and a `'resize'` handler to recalculate it on drag.

When we get the drag event, we only get its offset in pixels. We simply add that to its current offset. A `top` / `left` percentage based layout would be impossible to do (AFAIK) without a `'resize'` handler so we know where it's actually going to go. A `translate()`-based layout would work, but would require knowing the width of the element.

benmosher:
Ah, interesting points. I forgot about the `translate` percentage scope. That should just be something like `width of 1 column (%) * x column position * (total columns / item column width)`, right? A little weird, but doable. i.e. item at x=2, w=2 on a 4-wide grid is thus positioned at 100%.

Margins are a loose end. `calc([margin]px + [position]%)` seems to be viable as a `translate` unit (in Chrome beta, at time of writing), though it's pretty crazy looking.

It can be assumed the window won't resize during any grid item drag/resize event--right?--so any relevant pixel width info (be it for the item or the full grid) could be requested at the start of the mutation.

I'm not sure when I might dig into this, but I'll let you know if I get anywhere with it.

STRML:
Yeah, re: resizes while dragging that seems like a good assumption.

Margins will have to be calculated as percentages only. We can't rely on `calc()`.

[...]

PR #107, opened 2015-09-28
use requestAnimationFrame to detect any change in width

benmosher:
I just started using this strategy with some dynamic canvas containers to get 60FPS smooth resize, whether triggered by the window, or some other UI change.

In practice, I find it seems to be as performant as `window.on('resize', ...)`, but I haven't done any scientific tests. In theory, it makes sense it would work as well or better, because of the [point in the event loop at which `requestAnimationFrame` executes](http://www.html5rocks.com/en/tutorials/speed/animations/).

To test, try changing the `width` of the `#content` div to some explicit value, and watch the magic.

Note: I left the `resize` stuff there in case you want to [support <=IE9](http://caniuse.com/#feat=requestanimationframe). Could also use a polyfill, but I think perf may go down the tubes with `setInterval`.

Also: this solves the problem that prompted #104 for me. The pedantic purist in me would still rather have percentages all the way down, but this is a much lower-impact change that achieves roughly the same result.

### DIFF
--- a/lib/mixins/WidthListeningMixin.js
+++ b/lib/mixins/WidthListeningMixin.js
@@ -24,7 +24,11 @@ var WidthListeningMixin = {
 
   componentDidMount: function() {
     if (this.props.listenToWindowResize) {
-      window.addEventListener('resize', this.onWindowResize);
+      if (window.requestAnimationFrame) {
+        this.killListener = widthListener(this.getDOMNode(), this.onWidthChange);
+      } else {
+        window.addEventListener('resize', this.onWindowResize);
+      }
       // This is intentional. Once to properly set the breakpoint and resize the elements,
       // and again to compensate for any scrollbar that appeared because of the first step.
       this.onWindowResize();
@@ -32,8 +36,12 @@ var WidthListeningMixin = {
     }
   },
 
-  componentWillUnmount() {
-    window.removeEventListener('resize', this.onWindowResize);
+  componentWillUnmount: function() {
+    if (this.killListener) {
+      this.killListener();
+    } else {
+      window.removeEventListener('resize', this.onWindowResize);
+    }
   },
 
   /**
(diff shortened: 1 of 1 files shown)


STRML:
I don't mind the idea of debouncing resize events with RAF, but it definitely appears wasteful to me to touch the DOM every frame to measure size. 

While it carries some bulk, I've had really good experiences with https://github.com/marcj/css-element-queries/blob/master/src/ResizeSensor.js, which is really quite nice:

> no performance issues since it listens only on size changes of elements that have element query rules defined through css. Other element query polifills only listen on window.onresize which causes performance issues and allows only to detect changes via window.resize event and not inside layout changes like css3 animation, :hover, DOM changes etc.

benmosher:
Yeah, I hear that about touching the DOM. I was resistant to trying this technique as well, at first, and was only updating on `resize` and `transitionend` events.

I have been using this technique in practice in a number of places, though, and because of the place in the render cycle it is called, it's fast. As long as you only _read_, you're guaranteed to get already-laid-out values.

We use it to resize some canvas-based graphs while other UI elements transition in with CSS, so we get totally smooth transitions for both. :smile: 

Ideally, percentage-based widths are cleaner since the browser will handle them completely, but again, this works and is a much smaller, more tactical change.

Would you be willing to have this as an optional resize detection mode? My use case involves the canvas transitions I mentioned before, but I suspect many others won't need this level of resize sensitivity. I can imagine checking a flag property in addition to looking for RAF.

STRML:
Understood - I suppose I just have a visceral reaction to touching the DOM so often.

Let me try out the ResizeSensor approach once I finish the `next` branch. I am also thinking of just going 100% percentage-based which would fix all these problems.

benmosher:
closing because this is super old and I agree that percentages would be better anyway 😁

Issue #102, opened 2015-09-14
Grid Items not Re-Rendering when Component Props Update


AlanBreck:
I have an app that displays a grid of widgets using react-grid-layout. Each widget is dependent upon external data and is set up to display a loader until the data has been retrieved and the its props have been updated whereupon the component is re-rendered with the appropriate data. For some reason, though, the grid items aren't re-rendering after the props are updated. Any idea why?

Here's the block of code that calls the react-grid-layout component (see code comments):

```javascript
<ResponsiveReactGridLayout
	className="layout qtDashboard_widgets"
	layouts={{lg: col4, md: col4, sm: col2, xs: col2, xxs: col2}}
	breakpoints={{lg: 1200, md: 996, sm: 768, xs: 480, xxs: 0}}
	cols={{lg: 4, md: 4, sm: 2, xs: 2, xxs: 2}}
	onLayoutChange={this.onLayoutChange}
	isResizable={false} isDraggable={false}
	rowHeight={290} margin={[10, 10]}>
	{this.widgets().map( function ( widget, i ) {
		return (
			<div className="qtDashboard_widget_container" key={i}>
				{/*THIS BLOCK RE-RENDERS PROPERLY*/}				
				<DashboardWidget {...widget} appState={this.props.appState} />
			</div>
		);
	}, this )}
</ResponsiveReactGridLayout>
```

...and here's the DashboardWidget component render block:

```javascript
render: function() {
	return (
		<div className="qtDashboard_widget">
			{/*THIS BLOCK DOES NOT RE-RENDER*/}
			<SingleMetric metrics={this.props.appState.metrics} />
		</div>
	);
}
```

When I removed react-grid-layout from the equation, the re-renders took place as expected with the prop updates, but once re-inserted, the DashboardWidget component ceased re-rendering with prop updates.

Thanks, in advance, for your help!

mattquinn4:
You may need to use immutable data structures.  I'm guessing GridItem's shouldComponentUpdate is returning false when it compares props and nextProps because it is actually comparing a reference to the same object.  Have a look at this link for more info:

https://facebook.github.io/react/docs/advanced-performance.html#shouldcomponentupdate-in-action

AlanBreck:
Ahh... that makes sense. ImmutableJS solved it. Thanks so much!

benmosher:
[...]

PR #108, opened 2015-09-28
remove PureDeepRenderMixin

### LINKED ISSUE #105: DeepPureRenderMixin prevents GridItem children from rendering
I just ran into a wild issue where my grid would not update even though props were changing, likely similar to #102.

Digging into the code, I see the DeepPureRenderMixin is mixed into both the grid layout and the item wrapper.

At first blush, deep equality for props seems like it should always be correct, so I figured that must not be the problem.

After some digging in, though, in this case, but it turned out the problem was that the prop that was changing is always a `Promise`, and apparently `deepEqual(aPromise, bPromise)` is `true` for all `aPromise`, `bPromise`.

I have some code that uses frozen objects as `WeakMap` keys into Flux stores, as well, and these objects are not guaranteed to be distinct in value. I assume these would also fail to re-render on account of potentially being deeply equal.

Given all that, and after re-reading the [React docs explanation of PureRenderMixin](https://facebook.github.io/react/docs/pure-render-mixin.html), I would argue that as a library component, neither the grid layouts nor the grid item components should implement `shouldComponentUpdate`. The responsibility to implement this should instead fall on the importing application author.

The application author ought to be able to get the same perf improvements by implementing `shouldComponentUpdate` in the parent component of the GridLayout, and / or the child component of the GridItem.

If there are expensive regions of code in either the grid layout or item `render()` methods, instead of relying on `shouldComponentUpdate`, the component could cache the results of the expensive bits and only update these cached results if necessary, inside of `componentWillReceiveProps`.

I have a fork where I modified 0.8.5 to remove the `shouldComponentUpdate` implementation, and performance did not seem to significantly degrade. I can open a PR, but I though it would be good to discuss first.

benmosher:
Closes #105. Relates to #102.

Also: let me know if you want me to remove the other commit with changes to `build.sh`. Snuck in there because I couldn't run the build script tasks without it.


STRML:
Thanks for this.

PR #110, opened 2015-10-13
Pass in the element changed on resize and drag events.

runeli:
Passing the element changed allows to use the element in some other context. This requirement arises when using external libraries such as charts that depend on jQuery.


STRML:
Thanks, this was fixed in e292bfb.

PR #130, opened 2015-12-01
Add missing React import

mrblueblue:
When using this branch, I got a `React is undefined` error. Adding `React` as an import fixed this issue.

### DIFF
--- a/lib/components/ListensToWidth.jsx
+++ b/lib/components/ListensToWidth.jsx
@@ -1,5 +1,5 @@
 // @flow
-import { PropTypes, Component } from 'react';
+import React, { PropTypes, Component } from 'react';
 import ReactDOM from 'react-dom';
 
 type Props = {initialWidth: number, listenToWindowResize: boolean};


STRML:
On which line? This seems strange, React isn't referenced in this file.

mrblueblue:
The error is related to `Uncaught TypeError: Cannot read property 'onWidthChange' of undefined`, which is on line 77 in `build/ListensToWidth.jsx` or line 45 in `lib/ListensToWidth.jsx`.

It does seem strange but my theory is that any file with a React component needs to import require in `react` since it gets transpiled into `React.createClass` in the end.

STRML:
It certainly shouldn't be transpiling to `React.createClass`. Perhaps you have an outdated babel version? Wipe `node_modules` and retry.

mosse:
I'm seeing the same issue after wiping node_modules. Using babel 5.8.34. 

Thanks so much for publishing the beta. Super useful.

STRML:
Can you post some of the built `ListensToWidth.jsx`? I'm interested in how it could even compile this way.

mosse:
These are the troublesome lines. React is required as `_react`. Not sure where the error's coming from though.

```javascript
key: 'render',
      value: function render() {
        return React.createElement(ComposedComponent, _extends({}, this.props, { ref: 'child' }));
      }
```

STRML:
Ah you're 100% correct. There's even a plugin for it (https://www.npmjs.com/package/babel-plugin-react-require).

What a pain. Thanks for the PR!

PR #131, opened 2015-12-02
Allow container padding to be set independently of margin

seanich:
This is just a simple change to support setting the padding of the grid container independently of the spacing between items. In my particular use case this allows the grid items to align with the edges of the container without any hacks.

The padding can now be set via the `containerPadding` prop. It is an array of two numbers [x, y] just like `margin`. If the `containerPadding` prop is omitted the previous behaviour of using the margin settings as padding is used.

I've also changed the CSS `box-sizing` of grid elements in the examples since the extra width added by the borders becomes quite apparent when `containerPadding` is set to `[0,0]`.


slestang:
I really want this PR to be merge so I solve the conflict and updated the code to work with current master:
https://github.com/slestang/react-grid-layout/commit/9a502de6bea4c6a01a291f70f303c3f458580eb2

How we can help to have this merged?

enriquecaballero:
+1

seanich:
I've updated the code to work with the current master again. Sorry for taking a while to respond on this!

Let me know if there's anything else I can do to help have this merged :)

STRML:
:+1: Could you please remove the `dist` files from the PR?

seanich:
Done.

STRML:
screen shot 2016-08-08 at 9 48 43 pm

:shipit:

PR #132, opened 2015-12-07
modified 'synchronizeLayoutWithChildren' method to convert all uses o…

aweber1:
modified 'synchronizeLayoutWithChildren' method to convert all uses of child.key value to int


STRML:
Happy to accept this, but could you please remove the dist files?

aweber1:
Interesting, I see in the .gitignore that dist/* should be ignored, but it's not ignored in my local repo unless I add "/dist" to the .gitignore. Are you ok with that change to .gitignore?

STRML:
It's a bit misleading - I believe I did a `git add` on those files manually while ignoring the rest of the folder. I want `dist` to be available (for now) for users who want a browser build.

aweber1:
ok, i'll close the PR and submit a separate one without the dist files.

STRML:
No need for that, just:

```bash
git reset HEAD~1
git rm --cached dist/
git commit -m "message"
git push --force
```

PR #133, opened 2015-12-07
modified 'synchronizeLayoutWithChildren' method to convert all uses o…

aweber1:
modified 'synchronizeLayoutWithChildren' method to convert all uses of child.key value to int


STRML:
Thanks!

PR #134, opened 2015-12-08
changed LayoutItem.i property to be of type "string" instead of "number"

aweber1:
Not sure if you'll want this one or not. For the project within which I'm using react-grid-layout, I use strings (GUIDs) as the type for layout item ids, i.e. **LayoutItem.i**. Unfortunately, converting to integers isn't an option (for me at least), so it would be nice to retain the flexibility of using a string as the layout item key/id. 

I suspect there was likely a good reason behind converting to integer, but I'm hoping the use of strings won't be too abrasive.

cheers!

### DIFF
--- a/lib/GridItem.jsx
+++ b/lib/GridItem.jsx
@@ -47,7 +47,7 @@ export default class GridItem extends Component {
     },
 
     // ID is nice to have for callbacks
-    i: React.PropTypes.number.isRequired,
+    i: React.PropTypes.string.isRequired,
 
     // If true, item will be repositioned when x/y/w/h change
     moveOnStartChange: React.PropTypes.bool,
--- a/lib/ReactGridLayout.jsx
+++ b/lib/ReactGridLayout.jsx
@@ -194,14 +194,14 @@ class ReactGridLayout extends React.Component {
 
   /**
    * When dragging starts
-   * @param {Number} i Index of the child
+   * @param {String} i Id of the child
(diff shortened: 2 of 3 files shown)


STRML:
Yeah, this one makes sense. I think I may have gotten overzealous when refactoring and just assumed users would use numerical keys, which is not always going to be true.

Thanks for the PR.

Drecula:
Why type check this at all? Wouldn't it be better to let consumers of the component have the flexibility to use string or number?

STRML:
Technically React `key`s are always strings, and that's where it's used.

Drecula:
Understood, but I am suggesting that it would make use of the component more flexible if you handled type conversion for your internal representation as opposed to forcing consumers of it to convert their Ids to strings?

STRML:
I could, but honestly I don't like that kind of "magic". Just convert to a string on the way in.

Drecula:
Not to put too fine a point on it, but that means you push (bi-directional) type conversion back onto consumers of your component, so they have to implement 'magic' instead.

STRML:
No, my definition of "magic" is anything that's non-explicit. And silently converting your keys is non-explicit.

That said, I don't really care about it either way, it's such a minor point that we're both wasting time even talking about it. If you care enough & submit a PR, I'd be happy to merge it.

Drecula:
Cool. Thanks.

PR #137, opened 2015-12-18
enable passing className property to <GridItem/>

mrblueblue:
I want to be able to set a `className` property on the `<GridItem />`

It does seem to have `className` listed as an optional `propType`


STRML:
That `className` is intended for user only by placeholder elements. Otherwise, you will end up with a duplicated `className`: one on the `<GridItem>`, and the same one on its child.

Issue #125, opened 2015-11-04
Fractionary `transform: translate` in grid-items


queimadus:
Applying fractionary values on `transform: translate(x, y)` on a `grid-item` which has a child with a property that forces hardware acceleration will result in a fuzzy element in Chrome `46.0.2490.80 (64-bit)`.

This is presented in RGL when calculating the `grid-item`'s translate given the position in the layout and the grid's width.

Here is a simplified example (Retina displays will mitigate this issue; in non-retina displays the fuzzy text is evident):
https://jsfiddle.net/wtvdq4ep/

Would you consider adding a flag to round these values or make this the default behaviour?

slavik-m:
+1

STRML:
These are now rounded by default in `0.10`.

PR #141, opened 2016-01-14
Round size and position of GridItem in calcPosition()

eMaj-Fr:
I had some visual glitches because GridItem position and size could be floats. So the font were blurry.


queimadus:
This fixes #125

STRML:
:+1:

PR #149, opened 2016-02-09
Next

menelike:
I needed a working 0.14.x release of your great work in our current project. 
The changes (lots of refactoring) are massive, a few things like flow, docs, (what's up with the meteor files) have been left behind and a lot parts can be done much better (offsets on draggable). I am not interested in a permanent fork and therefore would like to see some efforts going back to your repository.

The showcase dev example should work fine.

I would appreciate a review since this is my first PR and I'm sure I missed some basics.

Cheers


STRML:
Impressive! I've been hoping someone would take up the mantle on this. I'll review it shortly.

NeXTs:
My hero

STRML:
What's the story with the new meteor entry point?

lib/GridItem.jsx
```diff
@@ -1,18 +1,16 @@
...
-import type {Position, Size} from './utils';
+import type {Position} from './utils';
 
 /**
  * An individual item within a ReactGridLayout.
  */
-export default class GridItem extends Component {
-
-  static propTypes = {
+const GridItem = React.createClass({
```
STRML: Any particular reason for switching back to `createClass`?

lib/GridItem.jsx
```diff
@@ -168,44 +193,38 @@ export default class GridItem extends Component {
...
    */
-  mixinDraggable(child: ReactElement, position: Position): ReactElement {
-    var start = typeof position.left === "string" ? undefined : {x: position.left, y: position.top};
+  mixinDraggable(child:ReactElement): ReactElement {
     return (
-      <Draggable
-        start={start}
-        moveOnStartChange={this.props.moveOnStartChange}
-        onStop={this.onDragHandler('onDragStop')}
+      <Draggable.DraggableCore
```
STRML: This was what I was hoping to do - <DraggableCore> was built intended for this module. Thanks for getting it going!

menelike:
I use this this library in a meteor project where I need to build a meteor-package/wrapper around everything. This additional build step enables me to use non minified and non optimized code to be used directly in the wrapper. This should actually be obsolete with meteor 1.3 (import/export support) or with an official npm release (meteor has an automatic npm wrapper package including browserify for this purpose).

I'm still looking for a solution where npm and a meteor package can coexist well. I'll push the wrapper repository by tomorrow, things should be self explanatory then.

STRML:
I see. So it will be safe to remove once we make a real release.

I'm going to check this out, run some tests, do a bit of linting and release. Thanks much for your contribution.

lib/GridItem.jsx
menelike: After digging deeper into the code I saw no win in using ES6 styled classes for all components. Imho the code looks much more clean with this solution. But I would definitely not insist on this. Moreover I don't know the impact on future flow type checks where this could be contra productive.

STRML: Understood. I rather like ES6 classes and I see the majority of new React libs being built in them. The effect is the same. It's useful with Flow as well for easily e.g. typechecking State.

It doesn't matter much - I may revert the createClass switch but the meat of this PR doesn't depend on that. Thanks.

menelike:
@STRML you saved me a lot of time in my current project, I'm really happy to help you out.

STRML:
This looks great so far in my initial testing.

Prepare for 0.10.0 to finally land - thanks for getting the ball rolling @menelike.

lib/GridItem.jsx
menelike: I could gather all needed changes in an updated PR within the next days.
Thanks for your advise on ES6 classes I think you have a very valid point. I will revert the changes by tomorrow.

STRML: No need. I'm going to merge this as-is and I have a commit prepared that will clean up those little bits, fix the Flow errors & update to Babel 6.

STRML:
:shipit:

PR #150, opened 2016-02-10
updated docs; fixed examples

menelike:
# updated breaking changes

- initialWidth renamed to width
- new HOC Provider logic on `WidthProvider`
- updated tests and docs to `WidthProvider` and component keys of type `String`

I do not like that we need offsetX and offsetY. This only exists because `DraggableCore` returns absolute positions which need to be synced to the grid layout. There must be a better way to solve this.

### DIFF
--- a/lib/ReactGridLayout.jsx
+++ b/lib/ReactGridLayout.jsx
@@ -49,7 +49,7 @@ export default class ReactGridLayout extends React.Component {
     verticalCompact: React.PropTypes.bool,
 
     // layout is an array of object with the format:
-    // {x: Number, y: Number, w: Number, h: Number, i: Number}
+    // {x: Number, y: Number, w: Number, h: Number, i: String}
     layout: function (props) {
       var layout = props.layout;
       // I hope you're setting the _grid property on the grid items
--- a/lib/ReactGridLayout.jsx
+++ b/lib/ReactGridLayout.jsx
@@ -250,10 +250,18 @@ export default class ReactGridLayout extends React.Component {
 
     // Set state
     this.setState({
-      layout: compact(layout, this.props.verticalCompact),
       activeDrag: null,
       oldDragItem: null
     });
+
+    // Set State and callback, but only if layout changed
+    const newLayout = compact(layout, this.props.verticalCompact);
+    if (!isEqual(newLayout, this.state.layout)) {
+      this.setState({
+        layout: newLayout
+      });
+      this.props.onLayoutChange(this.state.layout);
+    }
   }
 
   onResizeStart(i:string, w:number, h:number, {e, node}: ResizeEvent) {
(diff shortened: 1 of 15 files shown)


STRML:
:+1: Thanks a lot for another helpful contribution.

STRML:
Re: offsetX/offsetY, this was actually the whole point of 0.10 (and `<DraggableCore>`), and part of why 0.10 took so long. There weren't really any breaking changes in React 0.14 that caused the delay.

The real issue is that the original `<Draggable>` kept internal state of its own position, which makes it very easy to use, but difficult for more advanced use such as this. For that reason, I added the `moveOnStartChange` prop, which was very messy but worked for my purposes. 

But state should only be in one place. As I started expanding out this library, I found lots of small bugs caused by this duplication of state. That's when I started working on `<DraggableCore>`, which has no internal state whatsoever.

I think it's natural that the layout should have to manage the position of its children. It has to be able to move them when responsive breakpoints change or other items cause them to move in reaction. Moving them according to offsetX/Y in drag callbacks is just another part of managing the same state.

STRML:
I'm sorry, I am completely misunderstanding *which* offsetX and offsetY you're talking about.

You're correct about the ones on the layout itself. I'll work to remove the need.

lib/ReactGridLayout.jsx
```diff
@@ -250,10 +250,18 @@ export default class ReactGridLayout extends React.Component {
...
     // Set state
     this.setState({
-      layout: compact(layout, this.props.verticalCompact),
       activeDrag: null,
       oldDragItem: null
     });
+
+    // Set State and callback, but only if layout changed
+    const newLayout = compact(layout, this.props.verticalCompact);
+    if (!isEqual(newLayout, this.state.layout)) {
```
STRML: This actually won't fire because the compact/move functions modify the layout, they don't clone it.

Perhaps they should - I'll have to see how the perf stacks up.

STRML: Additionally, since `onLayoutChange` is *only* called on `Stop` handlers, this means that if the layout didn't happen to change in that very last movement, the callback won't fire.

I'll have to revert this piece.

menelike: @STRML your are absolutely right about that. Coming from Immutable States I always tend to make those mistakes where in place changes are made. Sorry for the inconvenience.

I already reverted this: 

https://github.com/menelike/react-grid-layout/commit/c5e6d8cb75bb82a8c0fc2aab06336e55e1eb2ae1

What would the proper way here, another PR? I do not want to pollute your PRs.

STRML: No worries - going to make some changes soon anyway, I'll put it in as part of that.

menelike:
@STRML Thanks for the great explanation on the `Draggable` history. Things are much clearer now.

PR #158, opened 2016-02-11
round size and position of GridItem in calcPosition (2)

eMaj-Fr:
follow up of #141 
Support when "resizing" and "dragging".


STRML:
:+1: Thanks!

PR #161, opened 2016-02-12
Fix container height calculation

geekjuice:
The container height was only taking the margin into consideration once rather than for each row.

Also added [HubSpot](http://www.hubspot.com) to the list of products that use RGL :+1: :beers:


STRML:
Thanks! Could you please exclude the package update & dist? I build them separately as part of a release commit.

geekjuice:
Ah, gotcha. Should be fixed now. :+1:

STRML:
:shipit: in `0.10.4`

geekjuice:
:boom: :rocket: :100: Thanks!

Issue #156, opened 2016-02-10
How should the parent pass in new layout(s)?


STRML:
Related to #98

This is an open issue for comment.

Currently, the following methods are used to determine whether or not to reload the layout:

RGL:

`!_.isEqual(this.props.layout, nextProps.layout)`

ResponsiveRGL:

``!_.isEqual(this.props.layouts, nextProps.layouts)`

This doesn't adequately cover some use issues, like:
- (#98) - If I reset the layout, to what was originally passed in, the props haven't changed, so the layout doesn't reset.

This is actually very similar to the problem that caused me to rewrite `<Draggable>` in the first place. It's difficult, in React, to enforce a state reset on children at some times while allowing them to manage their own state at other times.

Should users just simply grab a ref and `setState()` themselves? I don't know if I see a better option.

STRML:
Worst still, if you reset a layout by calling `setState()` on RRGL, it will only work once, because after that the new props always match the old props.

I don't want to move the check back to `!isEqual(this.state.layout, nextProps.layout)` because it seems that any errant rerender of the parent could cause the layout to reset. This means you'd have to always be certain to watch `onLayoutChanged` and `setState()` every time, which feels incredibly wasteful.

menelike:
Would this be solved if every passed prop and every state would be immutable?
Currently a lot of changes to layouts are made in place, but this sometimes makes debugging frustrating and you'll always need to look if your components update as they should.
As a result (surely with a performance loss) I always convert layouts to immutable (`Immutable.fromJS(layouts)`) and later pass RGL the reconverted mutable back (`layouts.toJS()`).

I use RGL on a Flux/Redux Immutable js stack. Maybe in long term it would be better to switch to Redux as well.

plandem:
+1 for redux

STRML:
Well, we definitely wouldn't use it in a library - not sure what you're suggesting. And Immutable is very slow and a large dependency.

We could clone the layout items when they come in. Since the shapes are regular we could do a very fast custom clone. That should help and we could start implementing a better SCU.

menelike:
@STRML redux without any explanation was indeed misleading. I may miss the point and this is total nonsense:

imho `shouldComponentUpdate` should only care about "do we need to render?" and never about "do we need to sync our states and then render?" Currently we have two sources of truth, props and states and we need to determine which to trust or if we merge. If we leave out the props we cut the API and if we leave out the states we loose complete control over the grid.

[...]

PR #162, opened 2016-02-12
Synchronize incorrect layout

geekjuice:
Something I noticed while testing this locally was that [this line](https://github.com/STRML/react-grid-layout/blob/master/lib/ReactGridLayout.jsx#L169) seems to synchronize against the current layout instead of the incoming layout. I would think that you would want to synchronize the next children with their layout and not the current layout. Changing it to `nextProps.layout` seems to fix a lot of issues I had with children not inheriting the correct layout from the grid component parent.

Is this intentional or could this possibly be a typo during the recent big update?

### DIFF
--- a/lib/ReactGridLayout.jsx
+++ b/lib/ReactGridLayout.jsx
@@ -166,7 +166,7 @@ export default class ReactGridLayout extends React.Component {
     // If children change, regenerate the layout.
     if (nextProps.children.length !== this.props.children.length) {
       this.setState({
-        layout: synchronizeLayoutWithChildren(this.state.layout, nextProps.children,
+        layout: synchronizeLayoutWithChildren(nextProps.layout, nextProps.children,
                                               nextProps.cols, nextProps.verticalCompact)
       });
     }


STRML:
This appears to be a mistake.

Could you explain how you are using it? This appears to be also related to the open-ended question in #156. For example, I could see the following flow:

* Developer passes an initial layout into RGL
* User drags items around. this.state.layout now !== this.props.layout.
* User triggers an action that creates a new layout item.
* If the grid were synchronized using `this.props.layout` as a base, all the changes the user made to the grid before adding the new item would be lost.

Perhaps we should be checking `this._pendingState` (or storing this ourselves because I don't think that's a stable API)?

geekjuice:
We have it setup so any change that is made to RGL triggers an action that creates a new layout that is passed back into RGL. So for the second step you described, once an item is dropped, we dispatch an event that takes the new layout, does stuff e.g. XHR, then pass it back to RGL as the next layout prop so `this.state.layout` should equal `this.props.layout`.

The reason I looked into this originally was we have multiple dashboard that can be swapped out (they all use RGL) and upon switching to a dashboard making the next dashboard layout incorrectly. Upon looking at React DevTools, the layout prop from `ReactGridLayout` didn't match up with the `_grid` props of it's children.

 :disappointed:  I know that may not have clarified too much ... but I'm still looking into the issue a bit myself, but the change in the PR did fix all the issues I was noticing with layouts being out of sync.

STRML:
Understood. So it seems we should be checking if both children & props have changed in that tick, and if they have, use the new layout.

Could you try: 

```js
componentWillReceiveProps(nextProps: Object) {
  let newLayout;
  // Allow parent to set layout directly.
  if (!isEqual(nextProps.layout, this.props.layout)) {
    newLayout = synchronizeLayoutWithChildren(nextProps.layout, nextProps.children,
                                                  nextProps.cols, nextProps.verticalCompact);
  }
  // If children change, regenerate the layout.
  else if (!nextProps.children.length !== this.props.children.length) {
    newLayout = synchronizeLayoutWithChildren(this.state.layout, nextProps.children,
                                              nextProps.cols, nextProps.verticalCompact);
  }
  if (newLayout) {
    this.setState({layout: newLayout});
    // Call back so we can store the layout
    this.props.onLayoutChange(newLayout);
  }
}
```

And let me know if it fixes your bug?

geekjuice:
:+1: :beers: Yup that worked!

STRML:
`0.10.6`

geekjuice:
:100: Thanks again!

Issue #93, opened 2015-08-25
Get the initial breakpoint?


tlrobinson:
It would be nice if there were a way to get the initial breakpoint, perhaps by calling `onBreakpointChange`    in one of the initial lifecycle methods.

STRML:
Since both the breakpoints and width are user-set I don't see this as an issue. Please reopen if you have more details.

sylv3r:
@STRML in responsive mode, they are not.

So how to get the initial breakpoint RGL is selecting based on its initial width ? I don't see any methods or callbacks helping here. `onBreakpointChange` is not triggered during initial rendering.

PR #174, opened 2016-02-25
Change initial width of WidthProvider to 0

renaudtertrais:
I had a inconsistency with RRGL with WidthProvider. It always trigger `onBreakpointChange()` when mounted except if the breakpoint deduced from the width of the container was the same as the one deduced from 1280px (initial width of the WidthProvider).

Not sur if it was intentional. But It seems more consistent now because it will always trigger `onBreakpointChange()` after being mounted.

What do you think @STRML ?

I think it solve #93 too.

### DIFF
--- a/lib/components/WidthProvider.jsx
+++ b/lib/components/WidthProvider.jsx
@@ -11,7 +11,8 @@ type State = {width: number};
 export default (ComposedComponent: ReactClass): ReactClass => class extends React.Component {
 
   state: State = {
-    width: 1280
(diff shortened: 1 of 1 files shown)


STRML:
I see why you would want to do this, but I don't think it should live here. 1280 is chosen as a sensible default to avoid redundant work.

I see two things that should be done:

1. Define an `initialWidth` property that WidthProvider can use so this is configurable by the end user, and,
2. `<ResponsiveReactGridLayout>` should always fire the `onBreakpointChange()` callback on mount.

Issue #187, opened 2016-03-18
ReactGridLayout requires key id for layout array


ro-savage:
Maybe I just missed this in the docos, but I wasn't able to find it.

When I created my layout array, I was required to pass a property for **_i_**, but the only mention I could find in the docos was for when using custom keys.

**WORKS**
```JS
import React from 'react'
import ReactGridLayout, { WidthProvider } from 'react-grid-layout'
const ReactGridLayoutWP = WidthProvider(ReactGridLayout)

const LoginPage = () => {
  const layout = [
    { x: 0, y: 0, w: 2, h: 3, i: '1' },
    { x: 2, y: 0, w: 4, h: 3, i: '2', static: true },
    { x: 6, y: 0, w: 2, h: 3, i: '3' },
  ]
  return (
    <ReactGridLayoutWP className="layout" rowHeight={30} layout={layout}>
      <div key={1}>1</div>
      <div key={2}>2 - Static</div>
      <div key={3}>3</div>
    </ReactGridLayoutWP>
  )
}

export default LoginPage
```

**DOESN'T WORK**
```JS
import React from 'react'
import ReactGridLayout, { WidthProvider } from 'react-grid-layout'
const ReactGridLayoutWP = WidthProvider(ReactGridLayout)

const LoginPage = () => {
  const layout = [
    { x: 0, y: 0, w: 2, h: 3 },
    { x: 2, y: 0, w: 4, h: 3, static: true },
    { x: 6, y: 0, w: 2, h: 3 },
  ]
  return (
    <ReactGridLayoutWP className="layout" rowHeight={30} layout={layout}>
      <div key={1}>1</div>
      <div key={2}>2 - Static</div>
      <div key={3}>3</div>
    </ReactGridLayoutWP>
  )
}

export default LoginPage
```

Am I missing something? Or does the documentation need to be updated?
Happy to make a PR to update docs if needed.

joayo:
+1

Sunakujira1:
+1

eMerzh:
on top of that it seems that if the 'i's are not contious (1 then 2 then 3....) the layout will be messed up....
might have a note in the doc as well :)

ro-savage:
This has been closed by PR #191 

Thanks for the feedback :)

PR #191, opened 2016-03-25
Update examples with id

ro-savage:
This is in reference to issue #187 

I assume this is expected behaviour, but it also could be a bug and the key id shouldn't be required?

**Changes:**
Update examples to show that the key and id are required.
Add the width in examples as it is required.
Add i to PropTypes
Move width to be at top of PropTypes as it is required.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,13 +85,13 @@ var ReactGridLayout = require('react-grid-layout');
 render: function() {
   // layout is an array of objects, see the demo for more complete usage
   var layout = [
-    {x: 0, y: 0, w: 1, h: 2},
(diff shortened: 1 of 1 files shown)


STRML:
Thanks. I agree that `i` should be autogenerated but until I have the time to fix that, it's good for the examples to be up to date.

bravo-kernel:
Priceless addition mr @ro-savage, this all makes sense now. :clap:

ro-savage:
@STRML If I get a chance, I will try take a look at implementing the auto-generating id.

Thanks for merging the PR and thanks for making this awesome plugin :)

PR #199, opened 2016-04-01
added rowHeightPercent property

voons:
For our product it is important that cells are square and the grid remains responsive, so we can not use a pixel value for the `rowHeight` property. To solve this, we added a `rowHeightPercent` property as an alternative, which overwrites the `rowHeight` behavior. Setting this to `100` results in a row height equal to the cells column width.

### DIFF
--- a/lib/GridItem.jsx
+++ b/lib/GridItem.jsx
@@ -24,7 +24,8 @@ export default class GridItem extends React.Component {
     // General grid attributes
     cols: PropTypes.number.isRequired,
     containerWidth: PropTypes.number.isRequired,
-    rowHeight: PropTypes.number.isRequired,
+    rowHeight: PropTypes.number,
+    rowHeightPercent: PropTypes.number,
     margin: PropTypes.array.isRequired,
     maxRows: PropTypes.number.isRequired,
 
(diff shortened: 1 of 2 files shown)


STRML:
Closing as part of v2 cleanup. This functionality is now achievable with the v2 API without any library changes:

```typescript
import ReactGridLayout, { useContainerWidth } from 'react-grid-layout';
import { calcGridColWidth } from 'react-grid-layout/core';

function MyGrid() {
  
  const cols = 12;
  const margin: [number, number] = [10, 10];
  const containerPadding: [number, number] = [10, 10];
  
  // Calculate column width, then derive rowHeight as a percentage
  const colWidth = calcGridColWidth({
    containerWidth: width,
    cols,
    margin,
    containerPadding,
    rowHeight: 0,
    maxRows: Infinity
  });
  
  // Square cells (100% of column width)
  const rowHeight = colWidth;
  
  // Or any aspect ratio: 75% = landscape, 133% = portrait
  // const rowHeight = colWidth * 0.75;
  
  return (
    <div ref={containerRef}>
      {mounted && (
        <ReactGridLayout
          width={width}
          gridConfig={{ cols, rowHeight, margin, containerPadding }}
        >
          {children}
        </ReactGridLayout>
      )}
    </div>
  );
}
```

Thanks for the feature request - the v2 hooks-based API makes this pattern straightforward without needing a dedicated prop.

PR #202, opened 2016-04-06
Bugfix: Draggable cancel selectors

RiiD:
Draggable cancel selectors should be separated with commas


STRML:
You're correct on this, but I'm thinking the solution should probably be to check if `this.props.cancel` is truthy, and if so join it, to avoid the dangling comma.

PR #203, opened 2016-04-06
Bugfix: Draggable cancel selectors

RiiD:
Cancel selectors should be separated with commas.


STRML:
No need for the array join - could be:

```js
const cancelSelector = ".react-resizable-handle" + (this.props.cancel ? "," + this.props.cancel : "");
```

STRML:
(by the way, no need to make a new PR, just amend and force-push so we can keep context)

RiiD:
Done.
It is my first pull request and I don't know all the stuff about it :)

STRML:
:+1: Thanks!

Issue #208, opened 2016-04-10
Rendering with new Breakpoints props do not refresh the layout


damienleroux:
Hello,

I was using react-grid-layout on a responsive page and I made this page configurable in live by user. it means a user can see how the layout is displayed depending on how he has configured breakpoints. It is some kind of layout builder using react grid layout if you know what I mean. If the code is worth it at the end, I'll be happy to share it on github. The problem is that when React-grid-layout is rendered, if i fill it with a new props "breakpoints", nothing refreshes. I am under the impression that the function used to refresh the layout is called only when initializing the state: https://github.com/STRML/react-grid-layout/blob/master/lib/ResponsiveReactGridLayout.jsx#L81-L83 and should be called in componentWillReceiveProps when receiving a new breakpoints props too.

What do you think?

Thank you for your advice

damienleroux:
I just want to add that the problem is identical for the props "cols"

STRML:
You're correct, there should be hooks to refresh the state based on this. A memoized selector like `reselect` may be a better way to construct these, rather than putting them in state at all.

damienleroux:
Yeah I agree, It is always tricky to keep state up to date with props info. I just find out another issue in `ResponsiveReactGridLayout` when injecting new width through props:

In `ResponsiveReactGridLayout.componentWillReceiveProps()`:
[...]

PR #217, opened 2016-04-16
Avoid unecessary responsive calculs when already done

damienleroux:
Hey!

I just had a `if else` statement that avoid to override calculs done by `onWidthChange()`. It settled one of the issue I had when sending a new width through `props`. This is related to [issue 208](https://github.com/STRML/react-grid-layout/issues/208). I'll add more code asap to correct the others problems I told you at the beginning in this issue.

Let me know what you think of it.


STRML:
This is good, thanks. In general, please don't commit built files - I only recalculate them at tagged releases to avoid diff churn.

PR #226, opened 2016-05-03
update react grid when filled with new cols or new breakpoints

damienleroux:
Hey!!

This settled [Rendering with new Breakpoints props do not refresh the layout](https://github.com/STRML/react-grid-layout/issues/208).

Let me know if you find anything wrong with this code. I test it through `npm run dev`. It looks fine for me.

Thank you


STRML:
Thanks! This is helpful. I hope some time soon to do a more proper fix and remove state entirely, instead memoizing calls that build up the layout - but until then this is a good stopgap.

STRML:
FYI, missed this in my (admittedly short) review, but the `cols` and `breakpoints` attributes are on `props`, not `state`. Flow caught this very quickly. If you're not using it, I highly recommend it.

STRML:
Changes are out in `0.12.3`.

PR #228, opened 2016-05-04
avoid having props.layouts.lg static overriden

damienleroux:
Hello

When changing the prop `static` on the layouts items I sent to `<ResponsiveReactGridLayout>`, I saw that the prop `static` was never take into account. So I made a modification to, for example, avoid having `props.layouts.lg` overriden by `this.state.layout` in case a new layouts was sent to `<ResponsiveReactGridLayout>` through `layouts`.

Thank you

### DIFF
--- a/lib/ResponsiveReactGridLayout.jsx
+++ b/lib/ResponsiveReactGridLayout.jsx
@@ -95,6 +95,7 @@ export default class ResponsiveReactGridLayout extends React.Component {
     || nextProps.breakpoint !== this.props.breakpoint
     || nextProps.breakpoints !== this.props.breakpoints
     || nextProps.cols !== this.props.cols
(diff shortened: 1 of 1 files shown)


STRML:
Yeah - this whole thing is kind of wrong, we shouldn't be modifying the incoming `layouts` prop at all. This simply avoids some of the damage but not all. I need to reconfigure this and treat `layouts` as immutable.

damienleroux:
Exactly! Event if my correction doesn't correct an Immutability issue, I think it will be good thing to let developers use a callback to have layout changes and let them give this new layout as props to your react grid component. Thank you for your fast answer 


Damien leroux

damienleroux:
too old. I close

PR #230, opened 2016-05-07
Substract paddings from node in WidthProvider

Niondir:
I have the problem that my container object has some padding which I like to keep. This leads to a wrong calculated width, since the content should be only rendered inside the padding. This code should Substrat any paddings.

### DIFF
--- a/lib/components/WidthProvider.jsx
+++ b/lib/components/WidthProvider.jsx
@@ -44,7 +44,14 @@ export default (ComposedComponent: ReactClass): ReactClass => class extends Reac
 
   onWindowResize = (_event: Event, cb: ?Function) => {
     const node = ReactDOM.findDOMNode(this);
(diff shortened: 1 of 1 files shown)


STRML:
Closing as part of v2 cleanup. This issue is fixed in v2's `useContainerWidth` hook, which uses `ResizeObserver`'s `contentRect.width` - this automatically excludes padding (measures the content box, not border box).

```typescript
import ReactGridLayout, { useContainerWidth } from 'react-grid-layout';

function MyGrid() {
  // width correctly excludes container padding
  
  return (
    <div ref={containerRef} style={{ padding: 20 }}>
      {mounted && <ReactGridLayout width={width}>...</ReactGridLayout>}
    </div>
  );
}
```

Thanks for reporting this issue - it's now handled correctly in the v2 API.

PR #231, opened 2016-05-10
updating rowHeight should update the grid layout

damienleroux:
A minor fix that is helpfull in my case. Thx


damienleroux:
too old. I close

akshaygoyal88:
can we set rowHeight to fixed? i.e. if items are resizable than also height should not increase to a particular height in px?

PR #232, opened 2016-05-10
Update  layouts[lastBreakpoint] at the right time

damienleroux:
I met a problem when changing the number of cols given as props. It recalculates the layout not using the layouts given in props but using the already modified layout present in `this.state`. I did a modification to have `findOrGenerateResponsiveLayout` using the `layouts` not yet modified. With this problem resolved, I can change the number of cols -1 then +1 and have the layout displayed as before. I have deliver a demo of this [builder](http://damienleroux.github.io/react-grid-layout-builder/demo/). Let me know what you think of it :)


damienleroux:
too old. I close

PR #249, opened 2016-06-01
Allow to override the compactItem function

nd0ut:
I was needed to make a custom compaction for my layout items: do the vertical compact only when item is at first or last column.

So now I'm just passed own `compactItem` that checks `item.x` and calls original `compactItem` with needed verticalLayout argument.


lib/ReactGridLayout.jsx
```diff
@@ -179,7 +179,7 @@ export default class ReactGridLayout extends React.Component {
     // We need to regenerate the layout.
     if (newLayoutBase) {
       const newLayout = synchronizeLayoutWithChildren(newLayoutBase, nextProps.children,
-                                                      nextProps.cols, nextProps.verticalCompact);
+                                                      nextProps.cols, nextProps.verticalCompact, this.props.compactItem);
```
STRML: This should be `nextProps.compactItem`

STRML:
I like where this is going. I'm thinking I really should work on making all of these functions overridable via defaultProps. It will take some passing around (probably the whole Props object) but that way we can open up all of these for modification.

Issue #266, opened 2016-06-19
How to move according to the handle ？


zhump:
How to move according to the handle ？
like this
http://gaearon.github.io/react-dnd/examples-customize-handles-and-previews.html

zhump:
draggableHandle 
3ks

jason-henriksen:
Please either un-close this or provide a working example.  I've put about 2 hours into making the draggableHandle property do anything and from my point of view it simply does not work.

Here's what I'm doing:  ( All of this is modifications of the examples )

  getDefaultProps() {
    return {
      className: "layout",
      cols: {lg: 12, md: 10, sm: 6, xs: 4, xxs: 2},
      rowHeight: 100,
      isDraggable: true,
      draggableHandle : 'gridItemDragHandle'
    };
  },

  createElement(el) {
    var removeStyle = {
      position: 'absolute',
      right: '2px',
      top: 0,
      cursor: 'pointer'
    };
    var moveStyle = {
      position: 'absolute',
      right: '17px',
      top: 0,
      cursor: 'pointer'
    };
    var i = el.add ? '+' : el.i;
    return (
      <div key={i} _grid={el}>
        {el.add ?
          <span className="add text" onClick={this.onAddItem} title="You can add an item by clicking here, too.">Add +</span>
        : <span className="text">{i}</span>}
        <span className="gridItemDragHandle" style={moveStyle}>[DRAG]</span>

        <span className="remove" style={removeStyle} onClick={this.onRemoveItem.bind(this, i)}>x</span>        
        <TestResizeChart width={this.state.width} height={this.state.height} {...this.props}/>        
      </div>
    );
  },

  onAddItem() {
    /*eslint no-console: 0*/
    console.log('adding', 'n' + this.state.newCounter);
    this.setState({
      // Add a new item. It must have a unique key!
      items: this.state.items.concat({
        i: 'n' + this.state.newCounter,
        x: this.state.items.length * 2 % (this.state.cols || 12),
        y: Infinity, // puts it at the bottom
        w: 2,
        h: 2
        ,
        draggableHandle : 'gridItemDragHandle'

      }),
      // Increment the counter to ensure key is always unique.
      newCounter: this.state.newCounter + 1
    });
  },


And in the css:

.gridItemDragHandle{
  border: solid 1px;
}


Please provide a working example.

STRML:
As the docs state:

> // A selector for the draggable handler

[...]

PR #268, opened 2016-06-20
Issue #266 - Expanded example text regarding the use of draggableHandle

jason-henriksen:
This change adds a tiny bit of commenting on the README.md to show that in the className you do not need the leading . character, however in the draggableHandle option you do need the leading dot.  Otherwise, the draggableHandle will not work.   

It seems kind of obvious in retrospect, but I lost over an hour trying to figure it out.  

This also adds to the static box example to create another grid item that can only be dragged by the handle.

Thank you for your patience.  
I am new to JavaScript and GitHub.  This is my first attempt at a pull request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -235,8 +235,13 @@ autoSize: ?boolean = true,
 cols: ?number = 12,
 
 // A selector that will not be draggable.
+// For example: draggableCancel:'.MyNonDraggableAreaClassName'
+// If you forget the leading . it will not work.
 draggableCancel: ?string = '',
+
 // A selector for the draggable handler
+// For example: draggableHandle:'.MyDragHandleClassName'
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -235,8 +235,13 @@ autoSize: ?boolean = true,
 cols: ?number = 12,
 
 // A selector that will not be draggable.
+// For example: draggableCancel:'.MyNonDraggableAreaClassName'
+// If you forget the leading . it will not work.
```
STRML: Could you mention instead that these are CSS selectors?

jason-henriksen:
OK.  I think this is what you asked for.  Not sure if I need to do anything to make that commit properly visible in the pull request.

STRML:
Great, thanks.

PR #273, opened 2016-06-29
use lodash to check object equality, prevents onLayoutChange loop

jakegibson:
Currently tries to do equality check on two objects (breakpoints, cols) this will always return false.  When using redux (or anything that would change props on layout change) onLayoutChange causes a loop due to always thinking it needs to run onWidthChange

### DIFF
--- a/lib/ResponsiveReactGridLayout.jsx
+++ b/lib/ResponsiveReactGridLayout.jsx
@@ -94,8 +94,8 @@ export default class ResponsiveReactGridLayout extends React.Component {
     if (
          nextProps.width != this.props.width
       || nextProps.breakpoint !== this.props.breakpoint
-      || nextProps.breakpoints !== this.props.breakpoints
-      || nextProps.cols !== this.props.cols
+      || !isEqual(nextProps.breakpoints, this.props.breakpoints)
+      || !isEqual(nextProps.cols, this.props.cols)
     ) {
       this.onWidthChange(nextProps);
     }


STRML:
I believe we've been back & forth on this a few times already. I need to do a little more investigation. This seems correct though; would rather do a more expensive deep equal than an unnecessary rerender.

jakegibson:
Either way doing the equality check with !== on objects is always going to return false so it would be better to do a proper check or take it out.  The rerender causes a loop if you are passing layouts in through props like you would with redux.

All the builds still seemed happy with the change.  Let me know if there is anything I can do to help it get merged.

STRML:
Reference equality won't always return false, if your data pipeline is changing the object at every render that's indicative of another issue. Your `Connect` call should return the same `layouts` object each time.

jakegibson:
Ah.. good point.  With redux you always return a new copy so unless you had a reducer that was only layouts it would just take more of a work around.  I still think it would be best to utilize isEqual with them being small, one level objects I assume the performance hit would be less than a rerender (haven't checked) and more flexible

STRML:
Yeah I agree that the perf hit is negligible. RGL is an expensive render (I've profiled it extensively on BitMEX) so it is best to avoid.

You shouldn't be returning a new copy from Redux - you might return a new object each time to merge into the parent's `props`, but the `layouts` object should be the same each time, unless you're doing something weirder I'm not aware of.

Anyway I think this is fine to merge. Thanks for the contribution.

STRML:
`0.12.7`

PR #284, opened 2016-07-11
Fix responsive cols + layout change

### LINKED ISSUE #281: Updating layout and cols in ResponsiveReactGridLayout uses old layout
It looks like right here https://github.com/STRML/react-grid-layout/blob/master/lib/ResponsiveReactGridLayout.jsx#L138 the new layout is overwritten with the current layout.

What is the purpose of this? Removing it appears to work for my usecase, but maybe there is some larger purpose? It looks like there are no tests (or at least unclear how to run them) so not sure how to validate the change.

Looking at #232 maybe this code should only be triggered when the breakpoints change?

maxkostow:
This is an attempt to fix #281.

### DIFF
--- a/lib/ResponsiveReactGridLayout.jsx
+++ b/lib/ResponsiveReactGridLayout.jsx
@@ -108,7 +108,7 @@ export default class ResponsiveReactGridLayout extends React.Component {
       // if one does not exist.
       const newLayout = findOrGenerateResponsiveLayout(
         nextProps.layouts, nextProps.breakpoints,
-        breakpoint, breakpoint, cols, nextProps.verticalLayout
+        breakpoint, breakpoint, cols, nextProps.verticalCompact
       );
       this.setState({layout: newLayout});
     }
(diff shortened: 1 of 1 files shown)


lib/ResponsiveReactGridLayout.jsx
```diff
@@ -125,27 +125,27 @@ export default class ResponsiveReactGridLayout extends React.Component {
...
     // Breakpoint change
     if (lastBreakpoint !== newBreakpoint || this.props.breakpoints !== breakpoints || this.props.cols !== cols) {
-
-      // Store the current layout
       const layouts = nextProps.layouts;
-      layouts[lastBreakpoint] = cloneLayout(this.state.layout);
 
-      // Find or generate a new one.
+      // Preserve the current layout if the current breakpoint is not present in the next layouts.
+      if (!(lastBreakpoint in layouts)) layouts[lastBreakpoint] = cloneLayout(this.state.layout);
```
STRML: This is fine for now - likely to remove entirely though. If you want to keep layout changes that the user makes, it should be mandatory to listen to `onLayoutChange` and update your model.

I'll likely go with a `WidthProvider`-style HOC for preserving the old hands-off behavior with the default core behavior being stateless.

STRML:
Thanks, this was definitely a bug. The saving back to `layouts` is certainly a bad choice as well and I need to do some thinking on how to fix this API.

PR #295, opened 2016-07-25
update instructions on dev page

nikolas:
`npm run dev-server` doesn't do anything, I think this should be `npm run dev`


STRML:
:+1: Thanks.

PR #296, opened 2016-07-25
Pluggable compact items

farmisen:
Makes the compactItem function overridable to allow custom packing strategies.


farmisen:
was supposed to be against my own fork - sorry for the noise.

PR #313, opened 2016-08-08
change the drag rendering code to reflect only the change movement

cnavrides:
This is modifying the calculation for dragging tiles. This implementation stores the original tile location, then calculates the delta change (from mouse movements) and uses that to then determine the new location. It has an offset on the click, which could potentially be a customization option for how much to offset the selected tile.

### DIFF
--- a/lib/GridItem.jsx
+++ b/lib/GridItem.jsx
@@ -136,8 +136,10 @@ export default class GridItem extends React.Component {
     }
 
     if (state && state.dragging) {
-      out.top = Math.round(state.dragging.top);
-      out.left = Math.round(state.dragging.left);
+      out.top = Math.round(out.top + state.dragging.deltaY);
+      out.left = Math.round(out.left + state.dragging.deltaX);
+      state.dragging.top = out.top;
+      state.dragging.left = out.left;
     }
 
     return out;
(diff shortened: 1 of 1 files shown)


STRML:
Could you provide some more background on the problem you're fixing with this?

cnavrides:
We had an issue where we were putting a grid layout (thanks for making this library, huge life saver for us!), inside a nested layout (4 levels deep with other items in some of the levels). This caused an issue where selecting an item to drag would then cause the calculation to be wrong and the item to shift in our grid one down and about 1.5 over. So moving tiles became odd as it wasn't fully pinned to the mouse. There is a todo comment ([here](https://github.com/STRML/react-grid-layout/blob/85f8b187c9ba413ddbbe4fdb90d713cffa3729f7/lib/GridItem.jsx#L293)) about it.

STRML:
Closing as part of v2 cleanup. The v2 rewrite implements delta-based drag positioning, which is exactly what this PR proposed.

In `GridItem.tsx`, drag position is now tracked using deltas:

```typescript
(e, { node, deltaX, deltaY }) => {
  let top = dragPositionRef.current.top + deltaY;
  let left = dragPositionRef.current.left + deltaX;
  // ...
  dragPositionRef.current = newPosition;
}
```

This should resolve nested grid dragging issues. If you still experience problems with nested grids in v2, please open a new issue with a reproduction.

Thanks for identifying this approach!

PR #323, opened 2016-08-24
Added possibility preserving aspect ratio in grid item

artem-kushal:
Added 2 new props in Grid Item such as  isSaveAspectRatio and aspectRatio. Grid item size(width, height) calculating with considering aspectRatio.

### DIFF
--- a/lib/GridItem.jsx
+++ b/lib/GridItem.jsx
@@ -59,6 +59,7 @@ export default class GridItem extends React.Component {
       if (typeof value !== 'number') return new Error('maxHeight not Number');
       if (value < props.h || value < props.minH) return new Error('maxHeight smaller than item height/minHeight');
     },
+    aspectRatio: PropTypes.number,
 
     // ID is nice to have for callbacks
     i: PropTypes.string.isRequired,
@@ -70,10 +71,12 @@ export default class GridItem extends React.Component {
     onResizeStop: PropTypes.func,
     onResizeStart: PropTypes.func,
     onResize: PropTypes.func,
+    changeWH: PropTypes.func,
 
     // Flags
     isDraggable: PropTypes.bool.isRequired,
     isResizable: PropTypes.bool.isRequired,
+    isSaveAspectRatio: PropTypes.bool.isRequired,
     static: PropTypes.bool,
 
     // Use CSS transforms instead of top/left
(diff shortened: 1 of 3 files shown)


mrsufgi:
Exactly what I was looking for! super!!

STRML:
This is actually an option on [Resizable](https://github.com/STRML/react-resizable/blob/master/lib/Resizable.jsx#L42). 

As part of RGLv2 I want to offer a lot more possibilities for overriding props (so things like this can easily be sent), including overriding the default `<Draggable>` and `<Resizable>` implementation if needed.

mrsufgi:
@STRML preserving the actual height and width (in pixels) is different then keeping the aspect ratio of columns and rows (which is done by Resizable)

martynasj:
Sounds good, when do you think of releasing v2?

maneetgoyal:
@STRML Are there any plans to merge this or support the locking of aspect ratio to a given value?

maneetgoyal:
@artem-kushal Do you have the time to re-sync this PR and resolve the merge conflicts? Thanks.

STRML:
Closing as part of v2 cleanup. Aspect ratio constraints can be implemented in v2 using the `onLayoutChange` callback:

```typescript
const ASPECT_RATIO = 1; // 1:1 = square, 0.5 = landscape, 2 = portrait

function MyGrid() {
  const [layout, setLayout] = useState(initialLayout);

  const handleLayoutChange = (newLayout) => {
    // Enforce aspect ratio: h = w * ratio
    const constrained = newLayout.map(item => ({
      ...item,
      h: Math.round(item.w * ASPECT_RATIO)
    }));
    setLayout(constrained);
  };

  return (
    <ReactGridLayout
      layout={layout}
      onLayoutChange={handleLayoutChange}
    >
      {children}
    </ReactGridLayout>
  );
}
```

You can also enforce this during resize specifically using `onResize`:

```typescript
onResize={(layout, oldItem, newItem, placeholder) => {
  // Lock height to width ratio during resize
  const constrainedLayout = layout.map(item =>
    item.i === newItem.i
      ? { ...item, h: Math.round(item.w * ASPECT_RATIO) }
      : item
  );
  setLayout(constrainedLayout);
}}
```

Thanks for the feature request!

PR #326, opened 2016-08-26
Add WidthProvider option to await nonzero width

WesHildebrandt:
This branch adds an option to WidthProvider to have it wait until the initially rendered &lt;div&gt; has a nonzero width before rendering the actual &lt;ComposedComponent&gt;.  There are also options controlling how often it polls for a nonzero width (default is 50 ms) and how long to poll before giving up (default is 2000 ms).  I needed this capability because I had RGLs (actually RRGLs) inside a tab component, and when switching tabs, WidthProvider saw an initial width of zero (since the new tab content area wasn't fully rendered yet) and the RRGL was rendered to that zero width.  So the RRGL was never visible since it never saw that the tab content width changed.

You'll also notice that I add an internal &lt;div&gt; as a child of the initially rendered &lt;div&gt;.  This prevents children of the &lt;ComposedComponent&gt; from being rendered before the width is properly set.  You may want to add that to the &lt;div&gt; in the measureBeforeMount case as well, although in that case the short interval between the first two render() calls makes it very unlikely that a user will observe the children before the second render.

### DIFF
--- a/lib/components/WidthProvider.jsx
+++ b/lib/components/WidthProvider.jsx
@@ -5,7 +5,8 @@ import ReactDOM from 'react-dom';
 
 type State = {
   mounted: boolean,
-  width: number
+  width: number,
+  widthFound: boolean,
 };
 
 /*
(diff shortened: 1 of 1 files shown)


STRML:
Hi Wes -

I appreciate this PR, but it's outside the scope of this module. `WidthProvider` is intended to be a very simple component that can be swapped out if you have more specific requirements.

If you package this component separately, I'll link to it in the README.

WesHildebrandt:
Are you open to including this as "NonzeroWidthProvider" alongside the current WidthProvider?  I could also add some documentation into the PR for you.  I had to do some digging into the RGL issues list to figure out how to use SizeMe (although that didn't work in my case), so I could contribute some additional documentation for SizeMe, also.

STRML:
Unfortunately that would add size to the build for a component that the vast majority of consumers won't need to use.

We could add it in a `plugins/` directory, but doing so would require reworking the build somewhat as to rename `lib/` to `src/` and populate `lib/` with built files, then `require()` them for Node/Webpack/Browserify users.

WesHildebrandt:
Okay, I'll look at creating a separate repo for this component and send you a link when it's ready.

Issue #257, opened 2016-06-08
onLayoutChange call too often


vinej:
I think that the onLayoutChange is called too often. The event is fired even if the size or the position have not changed. One of the problem is that a simple click inside a widget fired the onLayoutChange ?. Maybe a bug on my side!, but I tested a quick fix (see below) and it's seem to do the job. Do you think that you can do the modification into the main the right way ?

ReactGridLayout.prototype.onDragStop
...
=>   if (oldDragItem.x == x && oldDragItem.y == y) return;
       // Set state
       this.setState({
...
ReactGridLayout.prototype.onResizeStop
    ...
=>   if (oldResizeItem.w == w && oldResizeItem.h == h) return;
       // Set state
       this.setState({

Thanks

larrydahooster:
May this is related to a issue that I reported a couple of weeks ago: https://github.com/STRML/react-grid-layout/issues/237
 
onLayoutChange gets even fired if layout did not change.

vinej:
For your information, the modification I did in the previous example DO NOT work. I had to put the check into my code, because the widget was sometime freezing on the drag stop event.

I hope someone will find the right way to do this modification

Thanks

uneart:
I've got the same problem, running react-grid-layout version 0.12.4. `onLayoutChange` is triggered when I start dragging an item and leave it at the same position. Sometimes it is triggered twice when releasing the mouse. The behaviour seems to be somehow random.

In my application `onLayoutChange` calls a Redux action creator that dispatches a Firebase update. Upon being updated in Firebase, the data is fetched back to the Redux state and from there is plugged directly in the RGL props.

Could the problem be that when the props change again, `onLayoutChange` gets called again even if nothing changed?

vickyqiu:
Same problem running version 0.12.4.

luangch:
Same problem. I suspected that the `onDragStop` calls the `onLayoutChange` before layout have actually changed.

Another problem is the `onDragStop` returns a new modified layout state(not in view) when I drag and put an item at the same position, as it stated by @uneart. Even it has only one item in the layout, the `x` and `y` of the item state could be set to the number other than their original.

I have tested a fix in my fork and it is work. I don't know if it is a proper way to do it. Please review my PR.
`if (!isEqual(oldDragItem, l)) this.props.onLayoutChange(this.state.layout);`

STRML:
Merged #328. This will make it into the next release.

uttamsingapura:
[...]

PR #328, opened 2016-08-26
fix onLayoutChange call too often #257

luangch:
Prevent the onDragStop to call the onLayoutChange before it has done checking if an old item and a new one is not the same.

#257


STRML:
Thanks - appreciate this.

chaserstrong:
Does it really have been solved? In my version 0.16.6, the event will be fired yet even if a simple click was triggered inside a widget. And it will also been fired when component did mount which i think needn't been fired.

PR #335, opened 2016-08-31
Why do we pass props and state to measureBeforeMount <div /> in WidthProvider?

cp:
I'm curious about the reasoning is behind passing `this.props` and `this.state` to the `<div />` in `WidthProvider`.

So it seems to mount the grid components (via `this.props.children`), then remount them once `this.state.mounted` is true. This is presenting a problem for me since my grid components are stateful, and fire a async HTTP call on mount that results in a `setState(...)`. 

When the `setState(...)` occurs, the original instance has been unmounted so React fires an error in the console.

What are the reasons behind this, and the tradeoffs preventing it from being removed?

### DIFF
--- a/lib/components/WidthProvider.jsx
+++ b/lib/components/WidthProvider.jsx
@@ -48,7 +48,10 @@ export default (ComposedComponent: ReactClass): ReactClass => class extends Reac
   }
 
   render() {
-    if (this.props.measureBeforeMount && !this.state.mounted) return <div {...this.props} {...this.state} />;
+    if (this.props.measureBeforeMount && !this.state.mounted) {
+      return <div className={this.props.className} style={this.props.style} />;
+    }
+
     return <ComposedComponent {...this.props} {...this.state} />;
   }
 };


STRML:
Yeah, I can see how that would be an issue. The point was to pass `className`, `style`, etc., in case they modify how the div is laid-out.

It should be enough to simply pass any DOM properties.

cp:
@STRML Thanks for getting back to me! Do you think it'd be enough to pass only `className` and `style` as props to the div? I thought it could make sense to pass everything but children, but this presents a problem in React 15.3.1:

```
Warning: Unknown props `rowHeight`, `measureBeforeMount`, `onDragStop`, `onResizeStop`, `onResize`, `mounted` on <div> tag. Remove these props from the element. For details, see https://fb.me/react-unknown-prop
```

Thoughts?

STRML:
I think that's fine.

The point of `WidthProvider` is to be relatively minimal - if a user has more advanced requirements, I'd rather they just implement a new HOC than require code changes.

If you make that change I'll merge it. Thanks for opening this.

cp:
@STRML Awesome. I rebased with a new commit that includes the discussed functionality. Thanks for the help.

STRML:
Thanks!

Issue #356, opened 2016-09-26
synchronizeLayoutWithChildren cannot handle array within array


j-e-harris:
I was having some trouble with an error being returned stating that '_grid' was not a property of child.props in lin 376 of utils.js

This stemmed from creating elements in my JSX inside the ResponsiveGridLayout using the following method (example only):

`{objects.map(function(object, i){
        return <ObjectRow obj={object} key={i} />;
    })}`

Putting this into JSX generates a sub-array of children within the children variable of the synchronizeLayoutWithChildren() function in utils.js

I have found that flattening the children variable in this function resolves the issue.
Fix was to add the following code to line 362 of utils.js:

`children = [].concat.apply([], children); //flatten the JSX hierarchy`

Will organise a pull request for this fix if you need.

PR #357, opened 2016-09-27
Issue #356: Flatten the JSX hierarchy in children to stop '_grid' und…

j-e-harris:

example or test case.

…efined error

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -359,6 +359,7 @@ export function synchronizeLayoutWithChildren(initialLayout: Layout, children: A
   if (!Array.isArray(children)) {
     children = [children];
   }
+  children = [].concat.apply([],children); //flatten the JSX hierarchy
   initialLayout = initialLayout || [];
 
   // Generate one layout item per child.


STRML:
I'm not certain I understand the issue - can you explain?

j-e-harris:
What i found was that when the 'children' parameter passed into the synchronizeLayoutWithChildren function is an array containing sub-arrays, the function fails. Specifically this is because line 375 in lib/utils.js cannot handle a child element that is actually an array:
`if (process.env.NODE_ENV !== 'production' && child.props._grid) {`

This usually happens when you are inserting elements into JSX through a mechanism such as (example only):
`<ul>
        {this.props.results.map(function(result) {
           return <ListItemWrapper key={result.id} data={result}/>;
        })}
</ul>`

Here is a link to the React documentation that will assist:
https://facebook.github.io/react/docs/multiple-components.html#dynamic-children

Recommend you try something like this within a ResponsiveGridLayout and see the result.

Thanks for checking this out. Its actually causing some probs with a dashboard I'm building so it would be great to fix it.

STRML:
Yeah, `children` could actually be all kinds of things, like a keyed fragment.

Best to use `React.Children.map` instead. I'm going to fix this in a separate commit.

STRML:
This is fixed in https://github.com/STRML/react-grid-layout/commit/eac7e0817ab71813c413b1e22f387ccf66ec64ce which was released in `0.13.7`. Thanks for the contribution!

PR #387, opened 2016-10-24
fix typo in comment: leapflog -> leapfrog

nikolas:


STRML:
That's an unfortunate one!

PR #426, opened 2016-11-30
update lodash-related dependencies

nikolas:

example or test case.


STRML:
This is technically the same range - nothing we do depends a newer version than the lower bound of the range.

PR #437, opened 2016-12-13
Remove lib/ and Makefile from .npmignore

nikolas:
Ignoring these files prevents the ability to use custom branches of react-grid-layout via a GitHub url, like this: https://github.com/ccnmtl/juxtapose/blob/master/package.json#L30

Without `lib/` and `Makefile`, the custom build fails.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,10 +1,8 @@
-lib/
 test/
 examples/
 dist/
 yarn.lock
 .github
 *.config.js
-Makefile
 *.sh
 *.*.json


STRML:
This makes sense to me, but it also kind of breaks the whole point of `.npmignore`. Looking for a solution. I'm on vacation now so if you have the time, please let me know if you find a way to make this work properly while still keeping the bundle size down for npm consumers, which is 99% of regular use.

nikolas:
Yeah, I understand. I'll let you know if I find some best practices / what other npm packages do.

STRML:
Closing as part of v2 cleanup. This is now moot - v2 has a completely different package structure:

- `lib/` removed entirely (source is now `src/`, compiled to `dist/`)
- `Makefile` removed (now uses npm scripts)
- Build uses `tsup` for ESM/CJS output

If you need to use a fork as a dependency, the build command is now simply `npm run build`.

PR #438, opened 2016-12-13
Add .travis.yml

nikolas:
We don't have any automated tests yet for react-grid-layout, but we can
at the very least verify through travis that the package builds
without errors.


STRML:
Thanks!

PR #439, opened 2016-12-13
Limit webpack version to 1.13.x

nikolas:
webpack 1.14.0 seems to be causing problems: https://travis-ci.org/ccnmtl/juxtapose/builds/183712619#L801


nikolas:
nevermind, it was failing because of the missing webpack config file: https://github.com/STRML/react-grid-layout/pull/437

PR #441, opened 2016-12-14
Use isProduction const where available

nikolas:
Since this is already defined at the top of the file, we can use this variable in
synchronizeLayoutWithChildren instead of looking it up again.


STRML:
Thanks

PR #442, opened 2016-12-14
Set up unit tests with jest and add a few tests

nikolas:


STRML:
Excellent - thanks so much for getting the ball rolling on this.

PR #445, opened 2016-12-19
isStateless prop added

calmburst:
for the case, when you use redux to process and store layout - you pass new layout through props and internal state layout should not rewrite the one source of trooth

### DIFF
--- a/lib/ReactGridLayout.jsx
+++ b/lib/ReactGridLayout.jsx
@@ -81,6 +81,7 @@ export default class ReactGridLayout extends React.Component {
     // Flags
     //
     isDraggable: PropTypes.bool,
+    isStateless: PropTypes.bool,
     isResizable: PropTypes.bool,
     // Use CSS transforms instead of top/left
     useCSSTransforms: PropTypes.bool,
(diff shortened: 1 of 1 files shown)


HeroicAutodidact:
I would really love it if this was adapted into the core so I may stop using [this npm module](https://www.npmjs.com/package/react-grid-layout-stateless)

What needs to happen to accept this pull request?

calmburst:
I`ve solved the problem by mutating the react key of the element - so it renders from scratch with thу props, not internal state, althought it fires some warning from react, it works just fine

STRML:
Closing as part of v2 cleanup. v2 handles controlled/Redux integration better:

1. **Props sync to state**: When the `layout` prop changes, the component automatically syncs its internal state.

2. **`useGridLayout` hook** for fully controlled mode:

```typescript
import { useGridLayout } from 'react-grid-layout';

// Fully controlled with Redux
const layout = useSelector(state => state.gridLayout);
const dispatch = useDispatch();

const { onDragStart, onDrag, onDragStop, onResizeStart, onResize, onResizeStop } = useGridLayout({
  layout,
  cols: 12,
  width,
  onLayoutChange: (newLayout) => dispatch(setLayout(newLayout))
});
```

The hooks API lets you manage state entirely in Redux while using the library's drag/resize handlers.

Thanks for the feature request!

PR #460, opened 2017-01-17
Fixing drag handle property of GridItem.

ilacyero:

example or test case.


ilacyero:
I added this property in Grip properties:

## Grid layout properties
```
// Specifies a selector to be used as the handle that initiates drag.
handleItem: ?string
```
## Grid Item Props
```
// Specifies a selector to be used as the handle that initiates drag.
handle: ?string,
```

PR #466, opened 2017-01-24
add a class if an item is draggable

### LINKED ISSUE #465: Provide classes for dragging state
It's nice that I have classes to hook off of for resizing, but there is no class for if an element is draggable unless its already dragging.

chrisdrackett:
this matches with the resizable behavior and gives a hook for styling.

closes #465


STRML:
Fixed in 10050577a9e56fe251d5e09c6905a4c8b8878afa

PR #477, opened 2017-02-09
fix bug

johnsiu2016:
This is the first time I send pull request. I dun know any rules and I am developing in window so I dun know why there are some additional files changed. 

But I just want to point out the grid item default props: draggable and resizable do not work as what the document wrote

### DIFF
--- a/.jshintignore
+++ b/.jshintignore
@@ -1,0 +1,1 @@
+node_modules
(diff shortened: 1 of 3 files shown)


STRML:
These props do work as written, I fail to see the issue. Please explain in more detail what parameters fail to work as expected.

Regardless, I cannot accept this PR with the additional files, even if it were correct.

PR #480, opened 2017-02-11
Makefile reference to flow in local node_modules bin

milanlandaverde:
Instead of forcing flow to be global for a contributor 

😄


STRML:
Thanks - could you remove the extra line in `package.json`, `flow` is not the package you want, `flow-bin` is.

milanlandaverde:
@STRML updated!

STRML:
:shipit:

PR #489, opened 2017-02-16
Fix lint warnings

nikolas:
This fixes 4 flow warnings:

    warning  'componentName' is defined but never used  no-unused-vars


STRML:
:+1: Thanks!

Just for reference this can also be done in most cases by prepending an `_` (aka `_componentName`), as a way to indicate "I want to note for the future that this param is present, but I'm not going to use it."

PR #499, opened 2017-02-22
[BUG] CSS is case sensitive

dvoaviarison:
Not doing this can result into this error:

`Prop style passed to Widget. Has invalid keys msTransform`.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -312,7 +312,7 @@ export function setTransform({top, left, width, height}: Position): Object {
     transform: translate,
     WebkitTransform: translate,
     MozTransform: translate,
-    msTransform: translate,
+    MSTransform: translate,
     OTransform: translate,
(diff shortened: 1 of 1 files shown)


STRML:
Thanks!

dvoaviarison:
Wcm!
Do i need to wait for a release to get this change?

STRML:
Hang on - I clearly need to drink my coffee before coding, because `msTransform` is the correct usage. https://github.com/facebook/react/blob/469e68542b0c12d3c3a8ffbed664bd87fb6eadcf/src/renderers/dom/shared/CSSPropertyOperations.js#L49

dvoaviarison:
Oh, so this [GUY ](https://github.com/brigand/react-style-proptype/pull/6)is wrong!!!
PR'd there instead.

PR #539, opened 2017-04-18
Change PropTypes import to use prop-types instead of react. Fixes #538

### LINKED ISSUE #538: Deprecated PropTypes reference - needs to use prop-types package

- [x] Bug
- [ ] Feature Request
- [ ] Question

(Using react 15.5.4, react-grid-layout 0.14.4)

I am getting the following JS console warning:
`Warning: Accessing PropTypes via the main React package is deprecated. Use the prop-types package from npm instead.`

I believe it is originating from places like https://github.com/STRML/react-grid-layout/blob/master/lib/GridItem.jsx where PropTypes is imported from react instead of prop-types.

I believe this affects anyone using React 15.5 or greater.
See the note at the top of https://facebook.github.io/react/docs/typechecking-with-proptypes.html

If this is not the case, feel free to close. Thanks for your work on this project, love it so far.

inverts:

example or test case.

## **HEY - Full disclaimer, I did not test this! :)**
I just found all the places I could find where PropTypes was being imported from react and changed it to use prop-types, and added the dependency of course.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
   "dependencies": {
     "classnames": "^2.2.5",
     "lodash.isequal": "^4.0.0",
+    "prop-types": "^15.5.8",
     "react-draggable": "^2.1.1",
     "react-resizable": "^1.4.0"
   },
@@ -67,7 +68,6 @@
     "lodash": "^4.3.0",
     "pre-commit": "^1.1.2",
     "precommit-hook": "^3.0.0",
-    "prop-types": "^15.5.8",
     "react": "^15.0.0",
     "react-dom": "^15.0.0",
     "react-hot-loader": "^1.3.0",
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -67,6 +67,7 @@
     "lodash": "^4.3.0",
     "pre-commit": "^1.1.2",
     "precommit-hook": "^3.0.0",
+    "prop-types": "^15.5.8",
```
STRML: This will need to be a `dependency` as it is used at runtime.

STRML:
Thanks. There are two other missing references: WidthProvider and ResponsiveReactGridLayout are using `React.PropTypes`.

STRML:
Thanks!

STRML:
`0.14.5`.

Issue #560, opened 2017-05-19
Upgrade `react-grid-layout`


thasner:
- [ ] Bug
- [X] Feature Request
- [ ] Question

> Warning: Accessing PropTypes via the main React package is deprecated. Use the prop-types package from npm instead.

which is propagated from `react-grid-layout`, recently fixed in https://github.com/mzabriskie/react-draggable/issues/238

(possibly a duplicate of https://github.com/STRML/react-grid-layout/issues/556 )

Issue #556, opened 2017-05-12
Throwing "Warning: Failed propType"


melloc01:
- [X] Question

Hi, I am getting a warning when running the project. It'd be really cool if I have some help getting it out of there and understanding what is the incompatibility that was created.

The bin:
https://www.webpackbin.com/bins/-Kjwr_o7tLT1JAfz9ujE
if you open your console on the bin page you can see:

```
"Warning: Failed propType: Calling PropTypes validators directly is not supported by the `prop-types` package. Use `PropTypes.checkPropTypes()` to call them. Read more at http://fb.me/use-check-prop-types Check the render method of `BasicLayout`."
```

Versions:

react@15.0.2
react-dom@15.0.2
react-grid-layout@0.14.6

PR #561, opened 2017-05-19
Upgrade `react-grid-layout`

thasner:
Fixes https://github.com/STRML/react-grid-layout/issues/560

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "classnames": "^2.2.5",
     "lodash.isequal": "^4.0.0",
     "prop-types": "^15.5.8",
-    "react-draggable": "^2.1.1",
+    "react-draggable": "^2.2.6",
     "react-resizable": "^1.4.0"
   },
   "peerDependencies": {
--- a/yarn.lock
+++ b/yarn.lock
@@ -221,10 +221,6 @@ async@^2.1.2, async@^2.1.4:
   dependencies:
     lodash "^4.14.0"
 
-async@~0.2.6:
-  version "0.2.10"
-  resolved "https://registry.yarnpkg.com/async/-/async-0.2.10.tgz#b6bbe0b0674b9d719708ca38de8c237cb526c3d1"
-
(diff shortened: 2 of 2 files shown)


STRML:
Thanks!

thasner:
@STRML would it be possible to publish a new/patch version to npm?

STRML:
That's not required; the newest version of `react-resizable` matches the old constraints. You may want to run a new install or regenerate your own lockfile.

thasner:
the bit that's weird is I then need to add an artificial `"react-draggable": "^2.2.6"` dependency into my `package.json` just for this, and it'll eventually become stale.

It turns out this console warning isn't egregious, so I'll survive until you publish, but it still seems that publishing is the "right thing to do" from `react-grid-layout`'s perspective

STRML:
It's not, that's the whole point of these version ranges. Delete the entry
from your lockfile or shrinkwrap file and the package will update.

thasner:
I wasn't aware that npm would default to the newest compliant version.

NPM's [semver docs on `Caret Ranges`](https://docs.npmjs.com/misc/semver#caret-ranges-123-025-004) indicate that newer versions are "allowed". (There won't be a conflict if someone else requires a newer version in range)

Anyway, clearing `yarn.lock` and reinstalling did the trick. Thanks.

melloc01:
@STRML @thasner 

I'm still getting the errors as referenced also by my issue #556 
I've made a webpackbin https://www.webpackbin.com/bins/-Kjwr_o7tLT1JAfz9ujE that reproduces the error and you can check it out :+1: (you can check or change the versions on the `Cog` - top right)
I'm also having a perf issue because of this issue.

My lockfile:
```
react-draggable@^2.1.1, react-draggable@^2.2.6:
  version "2.2.6"

react-resizable@^1.4.0:
  version "1.7.0"

react-grid-layout@0.14.6:
  version "0.14.6"
```

PR #385, opened 2016-10-21
Add preventCollision option

nikolas:
This option prevents grid items from being moved when collided with. This
is useful functionality for the video sequencer I'm building with
react-grid-layout.

PR #564, opened 2017-05-23
Added isRearrangeable flag

EmrysMyrddin:
This option allows to prevent items to move away when dragging/resizing an element.
It simply add short circuit to `moveElement` and `onResize` when this option is enabled.

In combination with `verticalCompact={false}` and `maxRows` props, it allow me to build a dashboard with fixed height and width. `isRearrangeable` allows to bypass `maxRow` overflow when dragging and resizing elements since other elements can't be push to the bottom anymore.

Perhaps the prop name is not very good 🙂


lib/utils.js
```diff
@@ -217,12 +217,14 @@ export function getStatics(layout: Layout): Array<LayoutItem> {
  * @param  {Boolean}    [isUserAction] If true, designates that the item we're moving is
  *                                     being dragged/resized by the user.
  */
-export function moveElement(layout: Layout, l: LayoutItem, x: ?number, y: ?number, isUserAction: ?boolean): Layout {
+export function moveElement(layout: Layout, l: LayoutItem, x: ?number, y: ?number, isUserAction: ?boolean, isRearrangeable: ?boolean = true): Layout {
```
Jero786: Could be good to add some missing @param comments :)

Jero786:
@EmrysMyrddin  this PR work perfect!! In all cases! thanks! 

Work much, much better than this PR: https://github.com/STRML/react-grid-layout/pull/385
Si I think could be close the duplicated PR #385 and instead merge this better PR. WDYT?
@nikolas  @msiebert @STRML?

lib/utils.js
EmrysMyrddin: Yes of course, forgotten :-)

STRML:
This is pretty good, I like it.

Would you mind adding an example for it as well?

EmrysMyrddin:
@STRML Sure, is there any guide for generating exemples ? I see some scripts here, but I don't know how they work.

EmrysMyrddin:
@STRML I have added an example, say me if it suit you.

You should add a little bit of instructions for making exemples, in particular for being able to see the result of the exemple.

brentbaum:
Hey all,

We'd like to incorporate this feature in our next release (thanks @EmrysMyrddin for putting in the work!). What's necessary to push this PR the last mile and get it merged? I'm happy to take over if there are any remaining documentation tasks left.

STRML:
Merged and combined with the `preventCollision` PR (#629)

PR #606, opened 2017-08-18
Feature/dynamic units

zimoony:
This is an additional way of sizing the grid elements according to the viewport width/height of the browser window. Use 'vw'/'vh' as unit prop on ReactGridLayout or ResponsiveReactGridLayout. It might solves #104. SSR works. Readme related changes integrated and examples added.

Note that if 'vw'/'vh' is used a polyfill is needed for old browsers (see [here](https://www.w3schools.com/cssref/css_units.asp))

example or test case.


STRML:
Closing as part of v2 cleanup. Viewport-relative sizing is achievable with standard React patterns:

```typescript
function MyGrid() {
  const [viewportHeight, setViewportHeight] = useState(window.innerHeight);
  
  useEffect(() => {
    const handleResize = () => setViewportHeight(window.innerHeight);
    window.addEventListener('resize', handleResize);
    return () => window.removeEventListener('resize', handleResize);
  }, []);
  
  // 10vh row height
  const rowHeight = viewportHeight * 0.1;
  
  return (
    <div ref={containerRef}>
      {mounted && (
        <ReactGridLayout 
          width={width} 
          gridConfig={{ rowHeight }}
        />
      )}
    </div>
  );
}
```

This keeps viewport logic in userland where it belongs, rather than adding vw/vh parsing to the library.

Thanks for the contribution!

PR #625, opened 2017-09-13
add timeout hack; ref. https://stackoverflow.com/questions/26556436/r…

thisiscam:
This temporary hack fixes a bug that appeared on Safari. 

Specifically, on Safari the stylesheet is rendered slower/later than componentDidMount method. Causing Safari to return incorrect offsetWidth if the parent uses CSS for styling (also see https://stackoverflow.com/questions/3702246/safari-is-returning-the-wrong-width-value)

The hack came from this SO post:  https://stackoverflow.com/questions/26556436/

### DIFF
--- a/lib/components/WidthProvider.jsx
+++ b/lib/components/WidthProvider.jsx
@@ -43,7 +43,9 @@ const WidthProvider: ProviderT = (ComposedComponent) => class extends React.Comp
     // Call to properly set the breakpoint and resize the elements.
     // Note that if you're doing a full-width element, this can get a little wonky if a scrollbar
     // appears because of the grid. In that case, fire your own resize event, or set `overflow: scroll` on your body.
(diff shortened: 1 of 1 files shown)


STRML:
`<WidthProvider>` is only an example; I expect people who require other functionality to copy the component and modify it.

That said, this is a pretty ugly hack; I'd much rather not put it in core. Thanks for referencing the issue though, and hopefully this shows up for anyone who runs into it later on.

ylacaute:
Why do you say it is ugly hack ?
I am mainly a backend developper so my question is real. From what I understand of frontend tech : since browsers have only one thread to calculate AND to display, you have to live with that and use this kind of hack. I would even think : use it each time you can in order to simulate at the maximum on multi-thread system. Am I wrong in this way of thinking ?

Issue #641, opened 2017-10-10
Compaction doesn't work as expected


andrewhathaway:
Thanks for submitting an issue to RGL!

Please mark the type of this issue:

- [x] Bug
- [ ] Feature Request
- [ ] Question

I have a project currently using [this](https://www.npmjs.com/package/react-grid-layout-horizontal-compact) package due to the previous versions of RGL not support horizontal compaction. 

<img width="1117" alt="start" src="https://user-images.githubusercontent.com/1188830/31384166-a281e50e-adb5-11e7-8401-a69f951079ad.png">

When using this version, the compaction happens in way in which space is not wasted. Shown below:

<img width="760" alt="desired" src="https://user-images.githubusercontent.com/1188830/31384169-ab138a60-adb5-11e7-8b23-c1cbc5cb9cd4.png">

When using RGL `0.16.0`, the compaction is not as desired, and just lines up against one side. I've also found that is switches between left and right, randomly when state has been updated. Shown below:

<img width="784" alt="actual" src="https://user-images.githubusercontent.com/1188830/31384174-aeffbcde-adb5-11e7-8f03-753db8abe365.png">

I'd prefer it if compaction was just done like this, and blocks fit in to where they could?

Thanks.

larrydahooster:
I can confirm it. The compacting seems to be broken after upgrade to 16.0.0 or it is a breaking change. I haven't played with

> compactType: ?('vertical' | 'horizontal') = 'vertical';
> preventCollision: ?boolean = false;

parameters yet, but the behaviour with default params seems broken in my opinion.

0.15.3 
![oct-16-2017 17-31-48](https://user-images.githubusercontent.com/4989851/31621621-7b576848-b29a-11e7-9998-731560c11029.gif)

0.16.0
![oct-16-2017 17-49-05](https://user-images.githubusercontent.com/4989851/31621694-a1bd23ba-b29a-11e7-8e2d-f15713810140.gif)

larrydahooster:
I could also easily reproduce weird behaviour in the basic demo

![oct-16-2017 18-04-18](https://user-images.githubusercontent.com/4989851/31622407-9c7dde56-b29c-11e7-9a8a-a3d24a805991.gif)

torkelo:
I am so seeing this. Feel strange that items flow up through full width panels

andrewhathaway:
I just want it all to fit nicely together and work out where to position things, making it a nice grid of things, rather than a vertical line!

[...]

PR #650, opened 2017-10-20
fix(compaction/moveElement): Fixes vertical & horizontal compaction & collision movement

torkelo:
Noticed the same behavior that @larrydahooster  reported here https://github.com/STRML/react-grid-layout/issues/641#issuecomment-336935304 

- compactItem is currently used to fix collisions but this is not done recursively. So when moving items down it caused items below to jump through items. I had to change so that this function takes the layout as well so it can recursively fix collisions when moving elements out of the way.
- moveElementElementAwayFromCollissions create a fakeItem the y pos of
  this fake item was not set correctly and cause panels to jump above
  intermediate items.

- created unit tests that replicates the scenarios that the code changes fixes.


lib/utils.js
```diff
@@ -155,7 +178,7 @@ export function compactItem(compareWith: Layout, l: LayoutItem, compactType: Com
     if (compactH) {
       l.x = collides.x + collides.w;
     } else {
-      l.y = collides.y + collides.h;
+      resolveCompactionCollisionByMovingItemDown(l, collides.y + collides.h, layout)
```
STRML: I think we'll have to create a similar function for horizontal compaction, yeah?

lib/utils.js
```diff
@@ -324,7 +347,7 @@ export function moveElementAwayFromCollision(layout: Layout, collidesWith: Layou
     // Make a mock item so we don't modify the item here, only modify in moveElement.
     const fakeItem: LayoutItem = {
       x: compactH  ? Math.max(collidesWith.x - itemToMove.w, 0) : itemToMove.x,
-      y: !compactH ? Math.max(collidesWith.y - itemToMove.h, 0) : itemToMove.y,
+      y: itemToMove.y,
```
STRML: Can you explain why this was changed - and not changed for `x`?

torkelo: Yea, think both might be wrong. This was changed when horizontal compaction was introduced, not sure why the compactH ?  condition was introduced. so left it for horizontal compaction as I have not tested that.

https://github.com/STRML/react-grid-layout/commit/9b1260d4c41eb900a1e853277c4711114369de0b#diff-50e3aa130a4f97a42ee2cf111c7b1d9dL301

I can test horizontal compaction and update PR so that it address issues for that mode as well.

STRML: It very well may be. That would be very helpful if you could give it a look and ensure horizontal compaction works as expected.

torkelo:
Updated PR with new tests for horizontal compaction and fixes for those issues as well. debated unifying the function resolveCompactionCollisionByMovingItemToTheRight and resolveCompactionCollisionByMovingItemDown to a single function that takes compactType but the code became a bit harder to read and reason about and the collision logic is tricky enough so opted for two functions that handles each scenario separately.

larrydahooster:
Hey @torkelo , I tested your fixes and apparently the noted misbehaviour that came up at change from 0.15.3 to 0.16.0 is also there with your fix. Please understand that I did not had the time to deeply dig into the calculations made. I will debug it in detail soon.

But what I don't understand is: With 0.15.3 to 0.16.0 a new horizontal compact was introduced. Those changes affected the vertical compact mode as well, it causes at least the posted misbehaviours.

Since no breaking changes have been announced yet, wouldn't it be better to revert the vertical compact mode into a state that exactly behaves as in 0.15.3 and find a cleaner way of integrating the vertical compact?

torkelo:
> Since no breaking changes have been announced yet, wouldn't it be better to revert the vertical compact mode into a state that exactly behaves as in 0.15.3 and find a cleaner way of integrating the vertical compact? I just have the feeling that things get more complex instead of easier.

I did test 0.15.3 and found many of the strange behaviors where compaction collision fix caused panels to jump over barriers (Do to compactItem not taking into account how moving elements can cause recursive collisions). 

So reverting to 0.15.3 would only fix the minor of the two root issues. 

Your gif is showing how moving an element to the right causes an element to the right to jump below a wide element that is under the element your moving. I fail to replicate that in the 0-showcase demo with either vertical or horizontal compaction. What demo is that from?

larrydahooster:
@torkelo I created an branch where I added an example: https://github.com/larrydahooster/react-grid-layout/tree/exampleSetup

It is based on your changes. I somehow can't open an PR to merge it into your fork. Hope it helps you

STRML:
@torkelo is correct that these are other bugs that have existed in the compaction for a while. I'm going to test this soon .

torkelo:
added test cases for the new scenario but not sure how I can fix it yet as it is caused by the sorting of the layout.

When the layout looks like this:
```
i: 1,  x: 0  y: 0  h:1  w:1
i: 2,  x: 1  y: 0  h:1  w:1
i: 3   x: 0  y: 1  h: 1  w:2
```

And we move element 1 over to the right it will push down element 2 so we end up with this array:
```
i: 1,  x: 1  y: 0  h:1  w:1
i: 2,  x: 1  y: 1  h:1  w:1
i: 3   x: 0  y: 1  h: 1  w:2
```

So element 3 and 2 are now colliding but before the collisions the moveElement function (and compact function) will sort the layout array by position so we end up with this array.

```
i: 1,  x: 1  y: 0  h:1  w:1
i: 3   x: 0  y: 1  h: 1  w:2
i: 2,  x: 1  y: 1  h:1  w:1
```

As you can see now element 3 is in the middle, which is correct according to the sorting rules (since it's x value is 0). And now given this order the collision is correctly resolved by moving element 2 down. 

Think the sorting needs to take into account the moved flag.

torkelo:
Think I managed to solve that remaining case. 

Also caused by a change introduced in the horizontal compaction commit:

https://github.com/STRML/react-grid-layout/commit/9b1260d4c41eb900a1e853277c4711114369de0b#diff-50e3aa130a4f97a42ee2cf111c7b1d9dR288

The horizontal compaction commit adds this line to the moveElement collision check look:

``` if (l.x > collision.x && l.x - collision.x > collision.w / 4) continue;```

which causes the collision between 2 and 3 to be ignored (so 3 is not moved down)

larrydahooster:
Thank you, I will do another round of tests later that day or latest tomorrow!

torkelo:
let me know if you need anything else (refactor / more tests)

STRML:
Yeah thanks. Intensely busy at the moment but if you can think of any more tests to add they would be appreciated.

STRML:
Thanks, I'll take a look at it today. We're actually migrating something
internal to Grafana this week.

STRML:
Merged; please checkout `master` and give me a :+1: before release.

torkelo:
ok, great I will!

torkelo:
So far I have not encountered any issues, so please release another version on npm.

torkelo:
When you have time please push a new version to npm

torkelo:
@STRML here is a demo of upcoming version of Grafana using react-layout-grid, https://www.youtube.com/watch?v=BC_YRNpqj5k

If you want mention in README that Grafana is using this component :)

STRML:
@torkelo Excellent! Thanks for the heads up.

PR #652, opened 2017-10-24
Added ability to use fixed size item

adrienpwd:
Hi guys,

We needed the ability to use items with a fixed size within the Grid.
So I added a FixedGridItem and I tweaked some utilities functions.

To use it you have to pass responsive={false} to the grid, and you set your items with a size in pixels.

See the example 13, fixed-grid

Let me know what you think of that.

Thank you

Adrien


STRML:
This diff changes 60 files, appears you set most from 644 to 755. Could you please revert that?

Issue #346, opened 2016-09-12
RGL v2 Megathread


STRML:
Everyone,

Thanks so much for your use of React-Grid-Layout and a special thank you to all of you who have opened pull requests.

I haven't been able to spend as much time on this repo as I had hoped, as it fulfills our rather humble layout needs at [BitMEX](https://www.bitmex.com) quite well. Yet I understand that some of you have more specific needs and I would like to address those.

There are a large number of open issues that generally follow the same pattern:
- Confusion about input options or callbacks, especially `layout` vs. `data-grid/_grid`,
- Confusion about mutation of internal `layouts`
- Issues when using custom components as grid children
- Lack of extension points

I would like to address these issues with a minor rewrite in RGL v2 (yeah - skipping v1, we've been in 0.x for two years).
### React-Grid-Layout v2

RGL v2 has the following list of design goals:
#### Major Changes:

The primary goal of RGL v2 is to expose a set of simple, overridable components to make custom behavior easier to implement. I'd like to allow creation of new compactors, or even new layout mechanisms (`top/left` vs `transform`, or even global CSS munging).
- [ ] Pluggable implementations:
  - Compactor
  - DOM layout (`(l: LayoutItem, props) => props`)
  - CSS transition support (#571)
  - Any others?
  - Could this even be per-item?
- [ ] One, and only one way of defining a grid: via input to an exposed `<GridItem>` component.
  - A `<GridItem>` takes a `layout` property with the normal (`{w: number, h: number...}`) shape.
  - A `<GridItem>` always renders a `<div>` child so we can be assured that styles are settable no matter the item's children.
  - This component's implementation should be overridable (by extending the class) with a well-defined API.
  - This should serve as the primary extension point for implementing custom behavior.
  - This may serve as a way to pass items from one grid to another - while React will internally reset the state of the component, which could be problematic, it should be relatively easy to simply build `<GridItem>`s by spec and transfer them from one grid to another in the parent's `render()`.
#### Minor Changes:
- Expose `utils` and other internal functions. Allow a new implementation to be passed into the layout. This could look something like:

``` js
import {utils} from 'react-grid-layout'
// ...
utils.somethingToOverride = function()...
<ReactGridLayout __utils={utils} ...
```

> Thoughts on this? Are other libraries doing something like this for custom behavior?
- Pass pre- and post- compaction data in callbacks - see #345 - or should we allow hooking into the compactor somehow?
- Docs: Define grid lifecycle clearly and customization points
[...]

PR #653, opened 2017-10-24
Added FixedGridItem

adrienpwd:
This pull request add the ability to use items with a fixed size.

When you pass the prop responsive={false} to the Grid it will use the new FixedGridItem instead of the normal GridItem.

You still use the same layout Object, and all dimensions are now in pixels.

See example number 13.


EvHaus:
@STRML I just want to add some more context to this PR.

@adrienpwd and I work as Cumul8 where we're doing some really cool things with `react-grid-layout`. We have a need for a grid that isn't responsive. In other words, we need a grid where the columns and rows can be specified in pixels. This functionality is similar to how the grid operates in Microsoft's Power BI (https://powerbi.microsoft.com/en-us/) web app.

We realize this is a pretty large change to the `react-grid-layout` library and we're not sure if Adrien's implementation 100% in line with your vision. However, we're hoping that with your guidance we can get this feature in without any breaking changes to other users of the library. We think this feature could be super useful to others looking for the same behavior.

Please let us know if you have any feedback. We're really keen on working with you to get this merged in.

STRML:
I appreciate the PR. There are some style issues (file mode changes, tabs) that need to be addressed, but in general I won't be accepting this kind of PR (yet). I really would like this to be in the realm of plugins, where the base RGL is nearly useless without bringing at least one implementation (this could be bundled with the base install to incorporate the functionality we have today). Given a robust API, the functionality you've proposed today could easily live in userland without having to be maintained by the base lib. 

I would very much like to get that in first (see #346) before this hits.

EvHaus:
Alright, we'll sit on this for now and will revisit when v2 lands.

STRML:
Closing as part of v2 cleanup. Fixed-pixel sizing is outside the scope of a grid layout library.

If you need fixed pixel sizes, you likely don't need a grid library - just use absolute positioning:

```typescript
<div style={{ position: 'relative' }}>
  <div style={{ position: 'absolute', left: 100, top: 50, width: 200, height: 150 }}>
    Widget
  </div>
</div>
```

If you want grid-based drag/drop with fixed-ish sizes, you can lock dimensions using min/max constraints:

```typescript
const layout = [
  { i: 'a', x: 0, y: 0, w: 2, h: 2, minW: 2, maxW: 2, minH: 2, maxH: 2 }
];
```

The grid's value is responsive, proportional layouts - fixed pixel sizing works against that paradigm.

Thanks for the contribution!

Issue #603, opened 2017-08-10
WidthProvider do not handle window scroll appearing/disappearing 


shablenko:
- [x] Bug

WidthProvider do not handle window scroll appearing/disappearing

PR #669, opened 2017-11-28
Use an iframe for resize detection in WidthProvider

### LINKED ISSUE #37: column widths don't automatically adjust when scrollbar (dis)appears
Pretty much what the issue title is. Playing around with your demos, I noticed that when the scrollbar appears or disappears, the items don't autoadjust their size.

To repro what I'm talking about, open your basic demo and move things into the first column until the scrollbar appears (assuming you don't have it already). You should get an effect like this: 
screenshot from 2015-03-16 14 15 20
If you then resize the window manually, they autoadjust the width and the effect goes away. If you then make the scrollbar disappear by moving things out of the first column, the reverse effect happens: 
screenshot from 2015-03-16 14 19 24
For reference, this is what it should look like:
screenshot from 2015-03-16 14 18 51

codepunkt:
This also detects resizes occuring because a vertical scrollbar is shown or hidden.

Fixes #37 and #603.

### DIFF
--- a/lib/components/WidthProvider.jsx
+++ b/lib/components/WidthProvider.jsx
@@ -19,7 +19,6 @@ type State = {
  */
 type ProviderT = (ComposedComponent: ReactComponentType<any>) => ReactComponentType<any>;
 const WidthProvider: ProviderT = (ComposedComponent) => class extends React.Component<Props, State> {
-
   static defaultProps = {
     measureBeforeMount: false
   };
@@ -35,26 +34,33 @@ const WidthProvider: ProviderT = (ComposedComponent) => class extends React.Comp
   };
 
   mounted: boolean = false;
+  iframe: ?HTMLIFrameElement = null;
 
   componentDidMount() {
     this.mounted = true;
 
-    window.addEventListener('resize', this.onWindowResize);
+    this.iframe.contentWindow.addEventListener('resize', this.onIframeResize);
     // Call to properly set the breakpoint and resize the elements.
     // Note that if you're doing a full-width element, this can get a little wonky if a scrollbar
     // appears because of the grid. In that case, fire your own resize event, or set `overflow: scroll` on your body.
-    this.onWindowResize();
+    this.onIframeResize();
   }
 
   componentWillUnmount() {
     this.mounted = false;
-    window.removeEventListener('resize', this.onWindowResize);
+    this.iframe.contentWindow.removeEventListener('resize', this.onIframeResize);
   }
 
-  onWindowResize = (_event: ?Event) => {
+  onIframeResize = (_event: ?Event) => {
     if (!this.mounted) return;
     const node = ReactDOM.findDOMNode(this); // Flow casts this to Text | Element
-    if (node instanceof HTMLElement) this.setState({width: node.offsetWidth});
+    if (node instanceof HTMLElement) {
+      this.setState({width: this.iframe.offsetWidth});
+    }
+  }
+
+  saveIframe = (iframe: HTMLIFrameElement) => {
+    this.iframe = iframe;
   }
 
   render() {
@@ -62,7 +68,22 @@ const WidthProvider: ProviderT = (ComposedComponent) => class extends React.Comp
       return <div className={this.props.className} style={this.props.style} />;
     }
 
-    return <ComposedComponent {...this.props} {...this.state} />;
+    return (
+      <span>
+        <iframe ref={this.saveIframe} style={{  
+          height: 0,
+          margin: 0,
+          padding: 0,
+          opacity: 0,
+          overflow: 'hidden',
+          borderWidth: 0,
+          position: 'absolute',
+          backgroundColor: 'transparent',
+          width: '100%'
+        }} />
+        <ComposedComponent {...this.props} {...this.state} />
+      </span>
+    );
   }
 };
 


codepunkt:
@TrySound We don't need a 2.5kb gzipped thing if all we need is an iframe with a few styles. Or do we?

STRML:
I'm not likely to accept this because of the (significant!) memory and cpu drain of using an iframe for this. 

I have been using this in production without issue:

```jsx
// @flow
import React from 'react';

type Dimensions = {width: number, height: number, ref: ?HTMLElement};

export default class ResizeSensor extends React.Component {

  props: {
    children?: ReactRenderable,
    className?: string,
    onResize: (dimensions: Dimensions) => any,
  };

  lastDimensions: Dimensions = {
    width: 0,
    height: 0,
    ref: null,
  };

  contract: HTMLElement;
  expand: HTMLElement;
  wrapper: HTMLElement;
  rafTimeout: number;

  componentDidMount() {
    this.resetTriggers();
  }

  resetTriggers() {
    const {contract, expand} = this;
    if (!contract || !expand) return;

    contract.scrollLeft      = contract.scrollWidth;
    contract.scrollTop       = contract.scrollHeight;
    // $FlowIgnore we know it has a firstChild
    const firstChildStyle    = expand.firstChild.style;
    firstChildStyle.width    = expand.offsetWidth + 1 + 'px';
    firstChildStyle.height   = expand.offsetHeight + 1 + 'px';
    expand.scrollLeft        = expand.scrollWidth;
    expand.scrollTop         = expand.scrollHeight;
  }

  onScroll = () => {
    if (this.rafTimeout) window.cancelAnimationFrame(this.rafTimeout);
    this.rafTimeout = window.requestAnimationFrame(() => {
      if (!this.wrapper) return; // Unmounted
      const dimensions = this.getDimensions();

      if (dimensions && this.haveDimensionsChanged(dimensions)) {
        this.lastDimensions = dimensions;
        this.props.onResize(dimensions);
        this.resetTriggers();
      }
    });
  };

  getDimensions() {
    const el = this.wrapper;
    if (el && el.lastChild) {
      const ref: HTMLElement = ((el.lastChild: any): HTMLElement);
      return {
        width: ref.offsetWidth,
        height: ref.offsetHeight,
        ref // children
      };
    }
    return this.lastDimensions;
  }

  haveDimensionsChanged(dimensions: Dimensions) {
    return dimensions.width !== this.lastDimensions.width || dimensions.height !== this.lastDimensions.height;
  }

  render() {
    // eslint-disable-next-line no-unused-vars
    const {onResize, ...rest} = this.props;
    const className = `${this.props.className || ''} resizesensor-wrapper`;
    return (
      <div {...rest} ref={(el) => (this.wrapper = el)} className={className}>
        {this.props.children}
        <div className="resizesensor-trigger">
          <div ref={(el) => (this.expand = el)} onScroll={this.onScroll}>
            <div />
          </div>
          <div className="resizeSensor-contract" ref={(el) => (this.contract = el)} onScroll={this.onScroll}>
            <div style={{width: '200%', height: '200%'}} />
          </div>
        </div>
      </div>
    );
  }
}
```

codepunkt:
@STRML If your solution works, you could as well add that one to `react-grid-layout`.

About the iframe performance drain:
- Do you have any numbers or articles that examine this critically?
- Under which circumstances does an iframe drain performance?
- Does the same performance drain happen with an object[type=text/html]?

STRML:
An iframe initializes a whole new document. You can only do about a dozen of those before it really starts to effect performance, but even one has significant memory overhead versus the sensor idea. I don't have numbers, but you could try it or google around.

Not sure re: an object but the same caveat should apply if it initializes a new document.

codepunkt:
I didn't find anything worthwhile.

STRML:
Any what? Any benchmarks? Any solutions? Any slowdowns?

codepunkt:
Any current benchmarks or articles with meaningful numbers about slowdown due to iframe usage.

Most stuff i've found is years old and says you shouldn't overuse them.

jboler:
I user a similar HOC to handle scrolling for a grid that depends on the window width:
```typescript
import React, { ComponentClass } from 'react';
import { Responsive, WidthProvider } from 'react-grid-layout';

function ScrollResizeProvider<P>(
  GridComponent: ComponentClass<P>
): ComponentClass<P> {
  return class ResizeProvider extends React.Component<P> {
    state = {
      width: 0,
    };

    componentDidMount() {
      window.addEventListener('scroll', this.handleResize);
    }

    componentWillUnmount() {
      window.removeEventListener('scroll', this.handleResize);
    }

    handleResize = () => {
      if (this.state.width !== document.body.clientWidth) {
        this.setState({ width: document.body.clientWidth });
        window.dispatchEvent(new Event('resize'));
      }
    };

    render() {
      return <GridComponent {...this.props} onResize={this.handleResize} />;
    }
  };
}

 const ResponsiveGridLayout = ScrollResizeProvider(WidthProvider(Responsive));

 export { ResponsiveGridLayout };
```

STRML:
Closing this PR as v2 uses `ResizeObserver` via the `useContainerWidth` hook, which is the modern standard for resize detection.

`ResizeObserver.contentRect.width` automatically handles scrollbar appearance/disappearance since it reports the content box dimensions (excluding scrollbars).

The iframe approach was clever for 2017 when ResizeObserver wasn't available, but it's no longer needed. Thanks for the contribution!

PR #671, opened 2017-12-02
updated the information

ubaidsworld:
The widget can be resized. From one of the demo you can see.

example or test case.


TrySound:
@ubaidsworld Right now resizing is allowed by right bottom corner. TODO is about another corners and maybe edges.

PR #678, opened 2017-12-10
Make resize corner style customizable

TrySound:


STRML:
Lgtm, that's far better than the svg anyway.

tonix-tuft:
How can we change the resize handle style? I couldn't find it in the README and in the examples. Thanks!

PR #682, opened 2017-12-11
Add prettier

TrySound:

example or test case.


STRML:
Well, that certainly takes the human factor out. If it passes the tests then :+1: from me, although I generally don't like the way these types of commits kill `git blame`.

PR #684, opened 2017-12-12
Drag api

TrySound:
Reference api is based on [reason like createRef](https://github.com/facebook/react/pull/11555)

### DIFF
--- a/index-dev.js
+++ b/index-dev.js
@@ -1,5 +1,8 @@
-module.exports = require('./lib/ReactGridLayout').default;
-module.exports.utils = require('./lib/utils');
-module.exports.Responsive = require('./lib/ResponsiveReactGridLayout').default;
-module.exports.Responsive.utils = require('./lib/responsiveUtils');
-module.exports.WidthProvider = require('./lib/components/WidthProvider').default;
+const utils = require("./lib/utils");
+
+module.exports = require("./lib/ReactGridLayout").default;
+module.exports.createDragApiRef = utils.createDragApiRef;
+module.exports.utils = utils;
+module.exports.Responsive = require("./lib/ResponsiveReactGridLayout").default;
+module.exports.Responsive.utils = require("./lib/responsiveUtils");
+module.exports.WidthProvider = require("./lib/components/WidthProvider").default;
--- a/index.js
+++ b/index.js
@@ -1,5 +1,8 @@
-module.exports = require('./build/ReactGridLayout').default;
-module.exports.utils = require('./build/utils');
-module.exports.Responsive = require('./build/ResponsiveReactGridLayout').default;
-module.exports.Responsive.utils = require('./build/responsiveUtils');
-module.exports.WidthProvider = require('./build/components/WidthProvider').default;
+const utils = require("./build/utils");
+
+module.exports = require("./build/ReactGridLayout").default;
+module.exports.createDragApiRef = utils.createDragApiRef;
+module.exports.utils = utils;
+module.exports.Responsive = require("./build/ResponsiveReactGridLayout").default;
+module.exports.Responsive.utils = require("./build/responsiveUtils");
+module.exports.WidthProvider = require("./build/components/WidthProvider").default;
--- a/index.js.flow
+++ b/index.js.flow
@@ -5,6 +5,9 @@ export { default } from './build/ReactGridLayout';
 export { default as Responsive } from './build/ResponsiveReactGridLayout';
 export { default as WidthProvider } from './build/components/WidthProvider';
 
+export type DragApiRefObject = utils.DragApiRefObject;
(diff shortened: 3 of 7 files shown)


TrySound:
Internals are quite messy yet. Gonna fix in following PRs.

TrySound:
Example

```js
import GridLayout, { createDragApiRef } from 'react-grid-layout';

class App extends React.Component {
  dragApi = createDragApiRef();

  onDrag = event => {
    if (this.dragApi.value) {
      const { x, y } = getPosition(event);
      if (inViewport(x, y)) {
        this.dragApi.value.dragIn({ i: 'id', w: 2, h: 2, node: event.target, position: { x, y }, event });
      } else {
        this.dragApi.value.dragOut({ position: { x, y }, event });
      }
    }
  };

  this.dragStop = event => {
    if (this.dragApi.value) {
      this.dragApi.value.stop({ position: getPosition(event), event });
    }
  };

  render() {
    return (
      <GridLayout dragApiRef={this.dragApi}>
        {...}
      </GridLayout>
    );
  }
}
```

STRML:
Thanks - can you explain more about the motivation behind this (versus just passing those values on callback)?

TrySound:
Maintaining the same props (passed to grid) separatly is hard. Changes in lib may break UX implicitly for developer. Also right now it's not posible to maintain placeholder.

TrySound:
Modeling events with swapping states is also shot in the foot.

foodforarabbit:
I'm using this and it looks pretty solid. 

Just a few notes for the example (as of right now):

position takes {top, left} as parameters (instead of x, y)

inViewPort should return true if (x, y) is within the thing you're dragging into (e.g. the react-grid-layout component)

getPosition should return event.clientX or event.clientY from a typical drag event.

dragOut does not need any parameters
stop only requires position parameter, make sure it matches the position for dragIn exactly (calculation-wise) or you'll see the element "jump" when you drop it in

to pass in arbitrary values to the grid item (e.g. if you're managing state based on the layoutOnChange callback) you'll have to modify dragIn to accept additional paramters, and add them to const l = { i, w, ...}. You'll also need to modify cloneLayoutItem to keep the attributes you're adding on clone. See my master branch for the changes.

I needed to throttle onDrag for performance, but then event wasn't persisting its values. The way I implemented it was to have two functions: one that only extracted and saved variables from event (e.g. x position, y position) and then that would call onDrag which was throttled to 100ms. onDrag would get the variables from wherever I saved them. 

I really like the dragApiRef / createRef pattern.

gordeevbr:
This looks like just exactly what I need right now.
Can I expect this to be merged merged in and released any time soon?

fairps:
Will begging and groveling get this merged any quicker ;-)  Could really use it right now

samiyaakhtar:
We're looking for this as well, any chance it will be merged?

samiyaakhtar:
@TrySound @foodforarabbit We're trying to use this drag-api and wondering if there's any plan to merge it into this master branch anytime soon. Could you provide us with an update on any reasons why this is on hold? It'll be super helpful, thanks!

foodforarabbit:
@samiyaakhtar Sorry, I am neither the repo maintainer nor the PR owner.. I simply forked the repo and added changes myself to support my exact needs, which may not necessarily support generic use cases. The simplest reason is probably that there are just more important features / fixes that need to be addressed first either in this repo, or another project.

NunoCardoso:
This is essential to create a custom placeholder that recalculates its proper place while dragging external objects to react-grid-layout. This is a must-have.

STRML:
I apologize for letting this sit so long. If someone is willing to resolve the conflicts & write some docs/tests for it, I'll merge it.

sidati:
Hi @STRML, please take a look #935 😄

sh3nan1gans:
FYI @TrySound - I think there's a bug in ReactGridLayout.onResizeStop - whenever an item is resized the onLayoutChange prop is never called. This is because "same" layout is being passed to this.onLayoutMaybeChanged. I suspect it's a mutation issue 

I'm just now digging into it so I'll let you know what I find.

BTW - thank you very much for this PR, I've been using it with great success.

durek1337:
Any results to expect here?

sanathks:
@STRML any update about this feature?

STRML:
No - conflicts still need to be resolved and tests written.

morgs32:
Ok I have an updated version of the PR here: https://github.com/STRML/react-grid-layout/pull/1247
I'll look into tests. I'm thinking to copy example 15 and to copy the "droppability" test.

ofekkazes:
Hi,

I would like this feature to be added as it suits my usage with the library.
Exposing the drag API can give customizability to the developers using the library.

I would use this to callback the original drag API but throttle it to increase performance.

PR #685, opened 2017-12-12
Reduce flow config

TrySound:

example or test case.

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -2,21 +2,11 @@
 0.61.0
 
 [ignore]
-.*test
-./build
+.*/node_modules/jsxhint/.*
 
 [include]
-./lib
-index.js
 
 [libs]
 interfaces/
 
 [options]
-suppress_comment=\\(.\\|\n\\)*\\s*\\$FlowFixMe.*
-suppress_comment=\\(.\\|\n\\)*\\s*\\$FlowBug.*
-suppress_comment=\\(.\\|\n\\)*\\s*\\$FlowIgnore.*
-suppress_comment=\\(.\\|\n\\)*\\s*\\$FlowNewLine.*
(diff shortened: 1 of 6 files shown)


.flowconfig
```diff
@@ -2,21 +2,11 @@
 0.61.0
 
 [ignore]
-.*test
-./build
+.*/node_modules/jsxhint/.*
```
STRML: JSXHint can just be removed from the devDependencies

.flowconfig
```diff
@@ -2,21 +2,11 @@
...
 
 [include]
-./lib
-index.js
 
 [libs]
 interfaces/
 
 [options]
-suppress_comment=\\(.\\|\n\\)*\\s*\\$FlowFixMe.*
```
STRML: I like keeping these around in case we need to use them later on; Flow is behaving well now but won't always

.flowconfig
```diff
@@ -2,21 +2,11 @@
...
-suppress_comment=\\(.\\|\n\\)*\\s*\\$FlowBug.*
-suppress_comment=\\(.\\|\n\\)*\\s*\\$FlowIgnore.*
-suppress_comment=\\(.\\|\n\\)*\\s*\\$FlowNewLine.*
-suppress_comment=\\(.\\|\n\\)*\\s*\\$FlowIssue
-esproposal.class_instance_fields=enable
```
STRML: Guess we don't use these anymore? Nice to see.

TrySound: They are enabled by default :) But I have a plan for them. Binding arrays with method names is a bit noisy. Keeping arrows in constructor is not a big deal.

.flowconfig
TrySound: FlowFixMe is works out of the box. Another errors I always try to solve.

PR #686, opened 2017-12-12
Reduce eslint config

TrySound:

example or test case.


STRML:
Will need a rebase on the other PR I just merged

lib/components/WidthProvider.jsx
```diff
@@ -21,7 +21,7 @@ type ProviderT = (
   ComposedComponent: ReactComponentType<any>
 ) => ReactComponentType<any>;
 const WidthProvider: ProviderT = ComposedComponent =>
-  class extends React.Component<Props, State> {
+  class WidthProvided extends React.Component<Props, State> {
```
STRML: Should be `WidthProvider`

TrySound: But there is already WidthProvider.

STRML: Ahhh, I misunderstood the name. Thanks

PR #705, opened 2018-01-16
NULL option for compactType in Readme.md

sunitJindal:
**What's changed?**
Added 'null' option in list of values for `compactType` in Readme.md.

**Why?**
Right now if `verticalCompact` option is used while creating a new grid, following warning is printed in console:
```
verticalCompact` on <ReactGridLayout> is deprecated and will be removed soon.
Use `compactType`: "horizontal" | "vertical" | null.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -254,7 +254,7 @@ draggableHandle: ?string = '',
 verticalCompact: ?boolean = true,
 
 // Compaction type.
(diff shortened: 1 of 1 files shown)


TrySound:
@sunitJindal `?` is already means maybe type which is `null` or `undefined`.

sunitJindal:
@TrySound my bad. Missed that. Thanks for pointing out.

PR #708, opened 2018-01-18
fix overridding onStart behaviour #707

### LINKED ISSUE #707: Overriding onStart method on react-draggable doesn't behave as expected
- [x] Bug
- [ ] Feature Request
- [ ] Question

My aim was to limit the drag action to a particular html element within the widget. So instead of dragging a GridItem by clicking anywhere, user has to click on an icon on the top right corner to drag an item. See gif below for intended behaviour:

react-grid-layout

I have tried `draggableHandle` and `draggableCancel` props without luck. After reading the docs of both react-grid-layout and react-draggable. I have decided to override the `onStart` callback of Draggable library so it returns false under certain conditions to stop drag action.

```
// Called when dragging starts. If `false` is returned any handler,
// the action will cancel.
onStart: DraggableEventHandler,
```
As per the lines 313-316 within GridItem. `onStart` callback of Draggable should be overridden by overridding `onDragStart` on ReactGridLayout.
https://github.com/STRML/react-grid-layout/blob/master/lib/GridItem.jsx#L313-L316
```      
<DraggableCore
        onStart={this.onDragHandler("onDragStart")}
        onDrag={this.onDragHandler("onDrag")}
        onStop={this.onDragHandler("onDragStop")}
```

I am initializing `ReactGridLayout` with the following props to override `onDragStart`.

```
				<ReactGridLayout
					layout={this.state.layout}
					onLayoutChange={this.onLayoutChange}
					{...this.props}
					onDragStart={(layout: Layout, oldItem: LayoutItem, newItem: LayoutItem,
					placeholder: LayoutItem, e: MouseEvent, element: HTMLElement) => {
						let domElement = e.target;
						if (_.isElement(domElement) && !domElement.classList.contains('move-widget')) {
							return false;
						}
						return true;
					}}
				>
					{this.generateDOM()}
				</ReactGridLayout>
```

However, above failed to work because there are **no return statements around the event callback function calls**. So below check within Draggable core fails. 
https://github.com/mzabriskie/react-draggable/blob/master/lib/DraggableCore.js#L271-L272

```
    // Call event handler. If it returns explicit false, cancel.
    log('calling', this.props.onStart);
    const shouldUpdate = this.props.onStart(e, coreEvent);
    if (shouldUpdate === false) return;
```

I think ReactGridLayout should return the result of overridden `onDragStart` callback so Draggable `onStart` behaviour can be manipulated directly. In order to achieve this:

1. `onDragHandler` function in `LibGridItem.jsx` **Line: 418** should be:
```
return handler.call(this, this.props.i, x, y, { e, node, newPosition });
```

2. `onDragStart` function in `ReactGridLayout.jsx` **Line: 334** should be:
```
return this.props.onDragStart(layout, l, l, null, e, node);
```

I'll be opening a PR along with this issue shortly. Sorry for the long write-up, but I could not get the webpackbin template working.

ersel:
Fixes #707

### DIFF
--- a/lib/GridItem.jsx
+++ b/lib/GridItem.jsx
@@ -415,7 +415,7 @@ export default class GridItem extends React.Component<Props, State> {
 
       const { x, y } = this.calcXY(newPosition.top, newPosition.left);
 
-      handler.call(this, this.props.i, x, y, { e, node, newPosition });
+      return handler.call(this, this.props.i, x, y, { e, node, newPosition });
     };
   }
 
--- a/lib/ReactGridLayout.jsx
+++ b/lib/ReactGridLayout.jsx
@@ -331,7 +331,7 @@ export default class ReactGridLayout extends React.Component<Props, State> {
       oldLayout: this.state.layout
     });
 
-    this.props.onDragStart(layout, l, l, null, e, node);
+    return this.props.onDragStart(layout, l, l, null, e, node);
   }
 
   /**


STRML:
Thanks.

And re: #707 keep exploring the `draggableHandle` option (see react-draggable itself). That should be a simpler solution for your problem.

ersel:
No worries, thanks. I got it working with draggableHandle in the end but I thought the `onDragStart` behaviour needed fixing 👍

PR #716, opened 2018-02-05
Slightly narrow flow definition for "style" prop

Jabher:
narrowing flow definition for style attribute to make Flow typings more narrowed:
Object -> CSSStyleDeclaration ([bundled Flow class](https://github.com/facebook/flow/blob/v0.64.0/lib/cssom.js#L71))

### DIFF
--- a/lib/GridItem.jsx
+++ b/lib/GridItem.jsx
@@ -43,7 +43,7 @@ type Props = {
   usePercentages?: boolean,
 
   className: string,
-  style?: Object,
+  style?: CSSStyleDeclaration,
   // Draggability
   cancel: string,
   handle: string,
--- a/lib/ReactGridLayout.jsx
+++ b/lib/ReactGridLayout.jsx
@@ -43,7 +43,7 @@ type State = {
 
 export type Props = {
   className: string,
-  style: Object,
+  style: CSSStyleDeclaration,
   width: number,
   autoSize: boolean,
   cols: number,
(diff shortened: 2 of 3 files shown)


STRML:
Great, didn't know about that builtin - thanks.

TrySound:
@Jabher @STRML And now you want to write `width` and `height` and others as strings? It's not nice in react.

TrySound:
Btw it's breaking change.

STRML:
Ah that is indeed an issue. Will revert this barring a better typedef that works w/React.

Jabher:
Weird. Sorry for that.
Will probably spend some time on investigation why definition is implemented in that way

Jabher:
Just... wow.

Flow just do not have constrainted definition for react style prop (in comparison to [TypeScript](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/v15/index.d.ts#L780)); they [are aware of that](https://github.com/facebook/flow/issues/5192) and looks like they do not plan to fix it.
I've also checked some css-in-js solutions with flow typings and still nothing...
Sorry again for that. 

It actually makes me feel sad that we still do not have nice typings for style prop.

TrySound:
@Jabher PR to flow?

Jabher:
Looks so. But I feel it can take significant amount of time, so definitely not today :)
Also it's more question of maintenance first of all, not initial seed.

TrySound:
@Jabher I'm sure it will be maintained similar to native css props. The problem often is lack of time for initial seed.

PR #719, opened 2018-02-06
fix(utils): re-render when children data-grid changes #718

### LINKED ISSUE #718: [BUG] data-grid prop change does not trigger a layout change
Hi,

I am trying to use the `data-grid` api, in combination with some undo-redo in my application (that makes the dashboard be a "controlled" component), but when I change the values, the dashboard does not re-render with the new ones.

I reproduced the issue in a simplified project here: https://www.webpackbin.com/bins/-L4gPpW8DlggWuGtaAKk
One can try to re-activate the prop `layout` on the main component and see that the undo/redo will work.

I am planning to submit a PR for that

micaste:
This fixes #718 , an issue with the data-grid api


cjtafoya:
Is there a reason this is still hanging out as an open PR that fixes a known issue? We're still working off a forked branch incorporating this PR -- would be great to get a response from @STRML or other repo maintainers.

STRML:
It's unclear whether or not this actually fixes the bug, and unclear why the change on L596 is required. Could you please explain @micaste and add a test?

jonmanzo:
Merge this in already!!!  It's not even my codebase and it's easy enough to see how the addition on line 596 is needed.  Could it change to iterate over each prop of the existing layout to check equality, sure, but it's arguable that the added iterations are not worth the expense.  Until I found this PR, I had to force new keys to get elements to re-render, which is nonsense.

makis-spy:
> Merge this in already!!! It's not even my codebase and it's easy enough to see how the addition on line 596 is needed. Could it change to iterate over each prop of the existing layout to check equality, sure, but it's arguable that the added iterations are not worth the expense. Until I found this PR, I had to force new keys to get elements to re-render, which is nonsense.

Thanks for this! Hack or no hack I got it working with the key approach. Saved me at least an hour.

STRML:
I really want to kill `data-grid`, but will wait until v2 to do so. I'll merge this.

mateuszadamiec-a4bee:
> I really want to kill `data-grid`, but will wait until v2 to do so. I'll merge this.
thanks ^^

PR #729, opened 2018-02-19
Fix unexpected behavior with verticalCompact enabled in version 16.02

### LINKED ISSUE #714: Unexpected behavior with verticalCompact  enabled in version 16.02
- [X ] Bug
- [ ] Feature Request
- [ ] Question

There is some weird behavior when moving widgets introduced in the new version.
I have verticalCompact enabled. In version 15.03 it works ok:

Version 15.03 [example](https://drive.google.com/file/d/1DFYsTWNpsYzoHd03D94Xngx-vGxpSUmA/view?usp=sharing)

But in version 16.02 when I want to move one element the other is unexpectedly taking free space and jumps to the top and stays there. See example:

Version 16.02 [example](https://drive.google.com/file/d/1HDoc_Pxbd9sz0ite7ETIYZ_QvL6WtxO6/view?usp=sharing)

alexanderzobnin:
This PR fixes #714 
The root cause of this is `isUserAction` flag in `moveElementAwayFromCollision()` function. As described in comments:
>If there is enough space above the collision to put this element, move it there.
**We only do this on the main collision** as this can get funky in cascades and cause
unwanted swapping behavior.

But `isUserAction` is always `true`, not only in the main collision, so we should reset it after the first call.
I added a test case for this, it represents layout similar to one described in issue.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ dist/*
 npm-debug.log
 build/
 .idea
+.vscode
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -445,6 +445,8 @@ export function moveElementAwayFromCollision(
   // We only do this on the main collision as this can get funky in cascades and cause
   // unwanted swapping behavior.
   if (isUserAction) {
+    // Reset isUserAction flag because we're not in the main collision more.
+    isUserAction = false;
     // Make a mock item so we don't modify the item here, only modify in moveElement.
     const fakeItem: LayoutItem = {
       x: compactH ? Math.max(collidesWith.x - itemToMove.w, 0) : itemToMove.x,
--- a/test/spec/utils-test.js
+++ b/test/spec/utils-test.js
@@ -265,6 +265,36 @@ describe("moveElement", () => {
       ]
     );
   });
+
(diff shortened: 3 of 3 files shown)


torkelo:
please merge this, it's causing a lot of bug reports in Grafana :)

STRML:
Hey, excellent, I've been trying to figure out what causes this. I appreciate the PR.

torkelo:
yes we are seeing the same, seems my collision fixes have been nullified by this commit, https://github.com/STRML/react-grid-layout/commit/15503084fb7b0af826427c8c0706901e5745a39f 

if I go back before that I get no movement bugs

alexanderzobnin:
@STRML We've found some regression. Actually, PR #650 fixes this bug, it was merged and at 0c790cba3c10d347547733a7b250717fd4583b9b all works well. But something goes wrong after 15503084fb7b0af826427c8c0706901e5745a39f commit. This bug appears again. It's hard to catch this because test cases were changed also in this commit. Try to run original tests from #650 with 15503084fb7b0af826427c8c0706901e5745a39f - there are 2 failed cases.

torkelo:
@alexanderzobnin I detailed my findings here, https://github.com/STRML/react-grid-layout/issues/739#issuecomment-368814683

PR #732, opened 2018-02-20
README: Adds Kibana to projects using React-Grid-Layout

varundbest:
Kibana :- 
https://github.com/elastic/kibana/blob/master/src/core_plugins/kibana/public/dashboard/grid/dashboard_grid.js


STRML:
Great, that's a high-profile add.

PR #733, opened 2018-02-21
Fix issue on componentDidMount when props are already in sync but state is not

jca41:


hdwatts:
Thanks for this PR, had the same issue, works great 👍

topched:
Having the exact same issue? Any idea if this fix is going to be merged or do we need to fork and make the updates ourselves?

jca41:
Any updates on this?

Issue #740, opened 2018-02-26
Feature Request: Toolbox


jhob:
It would be great if we could dynamically add/remove widgets from RGL by dragging them into a toolbox widget.

STRML:
Know anybody with the skills to :shipit:?

jhob:
I have a work in progress [in this fork](https://github.com/jhob/react-grid-layout), with a [demo viewable here](https://jhob.github.io/react-grid-layout/examples/14-toolbox.html).

I'm currently working through two pesky issues:
- Items can reappear after being moved into toolbox, occurs primarily on non-default responsive sizes. Can be resolved by skipping the recovery of a 1x1 widget within `synchronizeLayoutAndChildren`, called in `ReactGridLayout::componentWillReceiveProps`, although a tidier solution might exist.
- Switching between responsive breakpoints with items in the toolbox can create issues. In particular, items can exist both in the grid layout and in the toolbox. A responsive toolbox solution is in progress.

JLindwick:
I really love the repo, i'm simply trying to use this without the generate dom and random elements. I want to use the components I have in place of them and i can't figure out how to allow for ontake and onput and the toolbox item itself without the ._map and item.i ... sorry if its a noob question

pinesh:
> I really love the repo, i'm simply trying to use this without the generate dom and random elements. I want to use the components I have in place of them and i can't figure out how to allow for ontake and onput and the toolbox item itself without the ._map and item.i ... sorry if its a noob question

6 years later I have this exact challenge too!

PR #741, opened 2018-02-27
Toolbox example

jhob:
Working toward https://github.com/STRML/react-grid-layout/issues/740

This example does not touch the underlying component API, it simply shows how to create a dropbox with click interactions to remove/add widgets to the toolbox.

@STRML if you'd like a drag + drop version, I have that too [here](https://github.com/STRML/react-grid-layout/compare/master...jhob:feat/toolbox-drag).


STRML:
Cool. Dropping the dragging seems to really simplify the implementation. This is great and we can definitely use it going forward. Thanks!

Issue #739, opened 2018-02-26
Collision cascades cause rearrangement


STRML:
See https://github.com/STRML/react-grid-layout/pull/729#issuecomment-368576689

Imagine a collision like in this example:

![draggable-compact-cascade](https://user-images.githubusercontent.com/1197375/36684687-c520dffe-1ae6-11e8-847e-2d471be43371.gif)

When 9 intersects with 12, we don't move 12 down because it would immediately compact back up. *However*, the code actually does move 12, and by extension, 16 and 19, down temporarily, then compacts them back. You don't see this visually, but it can manifest in certain low-height items switching position.

It will take some work to write test cases for this to ensure we don't regress. Happily accepting PRs for failing test cases that demonstrate this behavior, so we can move forward on fixing.

torkelo:
seems my fixes in https://github.com/STRML/react-grid-layout/pull/650 are no longer effective after this commit, 

https://github.com/STRML/react-grid-layout/commit/15503084fb7b0af826427c8c0706901e5745a39f

STRML:
I refactored those changes and all the original test cases are passing, perhaps you can add another failing test case?

torkelo:
@STRML when you refactored the tests I think you changed what the tests actually tested, when I restore the tests to how they looked after my PR they fail in current master as they the scenarios they test for are not handled correctly again

torkelo:
your https://github.com/STRML/react-grid-layout/commit/15503084fb7b0af826427c8c0706901e5745a39f commit changes so much so it's' very hard to see the changes (the test file moved for example),

I managed to restore one important test (there are more) this is just an example:
https://github.com/grafana/react-grid-layout/commit/a5caad51bffad5ed213bea3b030dc1fd452989bb 

Those changes competently changes the scenario. If I restore the test it very accurately breaks (in current master) as it shows the buggy movement is now back. if I go back to the commit before https://github.com/STRML/react-grid-layout/commit/15503084fb7b0af826427c8c0706901e5745a39f  the test passes and all the buggy behavior is gone. 

The sad part is that that commit was done before 0.16.1 so there is no published version that works. I should have spotted this early but have been using yarn linked local version with a commit before yours so had not noticed this until after we started to get bug reports in Grafana v5 beta. 

[...]

PR #745, opened 2018-03-01
Fix rearrangement on compaction

STRML:
WIP

rgl-badmove

Issue is that 0 moves through 3, 3 doesn't move at all, so after the conflict is resolved 3 naturally moves to the top.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -190,9 +190,12 @@ function resolveCompactionCollision(
   // Go through each item we collide with.
   for (let i = itemIndex + 1; i < layout.length; i++) {
     const otherItem = layout[i];
-
     // Ignore static items
     if (otherItem.static) continue;
+
+    // Optimization: we can break early if we know we're past this el
+    // We can do this b/c it's a sorted layout
+    if (otherItem.y > (item.y + item.h)) break;
 
     if (collides(item, otherItem)) {
       resolveCompactionCollision(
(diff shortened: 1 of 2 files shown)


STRML:
Ping @torkelo this appears to be what you were talking about in https://github.com/STRML/react-grid-layout/issues/739#issuecomment-369144475

torkelo:
yes, this is the bug I was referring to. 

Was working before this commit. 
https://github.com/STRML/react-grid-layout/commit/15503084fb7b0af826427c8c0706901e5745a39f 

Any ideas?

STRML:
Can confirm #650 did not have this issue. Investigating where the refactor ended up making a logical change.

STRML:
@torkelo Please review - I went through your original changes line by line and figured out what was missed; part of the confusion laid around [an erroneous `+1` in the original commit](https://github.com/STRML/react-grid-layout/commit/15503084fb7b0af826427c8c0706901e5745a39f#diff-50e3aa130a4f97a42ee2cf111c7b1d9dL390) when doing a reverse collision.

I believe fixes the issue completely but please run & test.

torkelo:
thanks for fixing this! We discovered this late in Grafana v5 beta and did not have time to investigate a lot due to so many other issues with the v5 release so we just did a temporary solution of using fork with my original PR. But great that you found the reason for why the refactor broke things. Will switch back so Grafana uses the react-grid-layout official npm version now.

Issue #750, opened 2018-03-07
With no vertical compaction, can't drag past some boxes to top, and bugs resolving collisions


brian-eightsolutions:
- [x] Bug
- [ ] Feature Request
- [ ] Question

When vertical compaction is turned on, some boxes can't be dragged through. If you try to drag one tile through another to the top of the screen nothing happens.

Also, when moving a box in a way that will push another box down, the collision resolving logic sometimes pushes the box far too far down. (This one is mentioned by jackie-benowitz 
in another issue: https://github.com/STRML/react-grid-layout/issues/720).

The following example is from the example page "React-Grid-Layout Demo 11 - No Vertical Compacting (Free Movement)":
![no_compaction_collision_bug](https://user-images.githubusercontent.com/34519927/37065932-e7e80c7a-2157-11e8-91d1-a7de9d2756fc.gif)

brian-eightsolutions:
The one line fix I have submitted actually fixes both of the issues. That line (in `moveElementAwayFromCollision()`) is supposed to move the box away from the collision, but when vertical compaction is on, doesn't move it at all. This causes bugs further downstream when `resolveCompactionCollision()` runs.

FoxxMD:
Looking forward to a solution for this...

PR #766, opened 2018-03-22
Fix for bugs when no compaction is on. 

brian-eightsolutions:
Both of these fixes are for when there is no compaction. See https://github.com/STRML/react-grid-layout/issues/750

For `moveElementAwayFromCollision()`, the issue is that if there is no compaction, `compactH` and `compactV` are both false, and no collision is resolved at all. The solution to this is to apply the vertical collision logic if `compactH` is not true ie. in all other instances. This is valid as there are other instances in the code that make the same assumption.

For  `resolveCompactionCollision()`, the code was detecting an item's collision with itself, thus moving everything down too far. This isn't a noticable problem when there is compaction, as it will resolve all the items moving down too far afterwards by compacting them back up, but the logic is still incorrect. When compaction is off though, the tiles get pushed down too far and remain there, leaving empty blank space.

Thanks!
Brian

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -122,7 +122,7 @@ export function childrenEqual(a: ReactChildren, b: ReactChildren): boolean {
  * Given two layoutitems, check if they collide.
  */
 export function collides(l1: LayoutItem, l2: LayoutItem): boolean {
-  if (l1 === l2) return false; // same element
+  if (l1.i === l2.i) return false; // same element
   if (l1.x + l1.w <= l2.x) return false; // l1 is left of l2
   if (l1.x >= l2.x + l2.w) return false; // l1 is right of l2
   if (l1.y + l1.h <= l2.y) return false; // l1 is above l2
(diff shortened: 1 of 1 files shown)


lib/utils.js
```diff
@@ -192,6 +192,8 @@ function resolveCompactionCollision(
     const otherItem = layout[i];
     // Ignore static items
     if (otherItem.static) continue;
+    // Don't resolve collision if with itself
+    if (otherItem.i === item.i) continue;
```
STRML: This shouldn't be possible given that we iterate through the list starting past this item. Have you seen it compare with itself before?

lib/utils.js
```diff
@@ -441,7 +443,8 @@ export function moveElementAwayFromCollision(
   cols: number
 ): Layout {
   const compactH = compactType === "horizontal";
-  const compactV = compactType === "vertical";
+  // Compact vertically if not set to horizontal
+  const compactV = compactType !== 'horizontal';
```
STRML: Good call, we need to resolve the collision somehow.

STRML:
Thanks!

makis-spy:
Hi! Sorry to bring this up again but  were these "bugs" solved in a previous version ? I'm seeing the behavior in latest.

Thanks!

Issue #769, opened 2018-03-26
[Bug] onWidthChange is only called on breakpoint changes


species-gh:
Please mark the type of this issue:

- [x] Bug
- [ ] Feature Request
- [ ] Question

Hi,
I noticed that the callback `onWidthChange` is only called when there is a breakpoint change, but _not on every_ change of width, like the name suggests. Calling it on every width change makes this callback a lot more useful, since there already is a callback for breakpoint changes (`onBreakpointChange`).
This behavior can be observed in [this example](https://species-gh.github.io/react-grid-layout/examples/15-badOnWidthChange.html).

Greetings,
species

STRML:
Fixed in #770

sgordon3:
Can a new version be published to include this pull request?

dioscuroi:
Has this been published to npm? onWidthChange is still not being called unless breakpoint-change occurs..;;;

mankinchi:
No update on this?

PR #770, opened 2018-03-26
Fix onWidthChange only being called on breakpoint changes

species-gh:
This PR fixes issue #769


STRML:
Yeah, that's a bug. Thanks!

PR #773, opened 2018-03-26
Reuse types from lib instead of copying to build

TrySound:


STRML:
Yeah, that's a better idea. Thanks!

PR #774, opened 2018-03-27
Remove power assert and refactor tests with jest

TrySound:
power assert do not show anything helpful in addition to jest expect.
There are a few nice jest utilities like `expect.objectContaining` which overrides inlined ones.


STRML:
Cool -- nice one.

PR #775, opened 2018-03-28
Add babel runtime and replace es2015 with env

TrySound:
babel runtime allows to reuse the same code of helpers across files and packages


package.json
```diff
@@ -51,10 +52,12 @@
     "babel-plugin-react-transform": "^3.0.0",
     "babel-plugin-transform-react-constant-elements": "^6.5.0",
```
STRML: These should be removable (pulled in by the react preset)

TrySound: I see it here
https://github.com/STRML/react-grid-layout/blob/master/webpack-examples.config.js#L21-L22

TrySound:
Thoughts?

STRML:
What's the actual reduction in compiled size? I'd rather avoid the large babel-runtime dependency if possible.

TrySound:
react-grid-layout.min.js 77.47K

PR #776, opened 2018-03-28
Omit  prefix in component names

TrySound:
I think `React` prefix in every react component is not necessary


STRML:
I hesitate to accept this as it will break anyone deep importing any
components. Not that we support it, but the pros don't outweigh the cons.

TrySound:
I'm gonna propose flat bundles instead of transpiled files to have better control over size which users consume. We could start working on the next major release. What do you think?

STRML:
I would leave that up to the user's bundler. A flat bundle will also pull in optional modules like the ResponsiveReactGridLayout, and anything else we might choose to add. I don't see a good reason to pre-bundle.

TrySound:
The case is controllable treeshaking. So we could provide the only one and safe way to use the library with only named exports. I think this is better since user won't care about package file structure and won't need to choose how to use the library.

PR #790, opened 2018-04-09
fix static mode display bug

### LINKED ISSUE #789: different display between static and no-static

- [x] Bug
- [ ] Feature Request
- [ ] Question

( seems the template link was disabled )

When I set all elements `static` , it shows difference to when I set all `no-static`.

I check the source and found the reason. 
https://github.com/STRML/react-grid-layout/blob/5abce095edfd367d69a2efe8d21d0e7608610763/lib/utils.js#L269-L293

Imagine we have a array which **has collision between first element and last** .

```javascript
const layout = [
  { i; 1, x: 0, y: 0, w: 6, h: 4 },
  // ...
  // ...
  { i; 4, x: 0, y: 0, w: 6, h: 4 },
]
````

 In `static mode` the **first one** would change it's Y axis position because the `collidesWith`(line 273 & 287)  has include all elements, and `getFirstCollision` would return `true` on last element.

 But in `no-static mode` it is the **last one** which change it's Y postion because when check the first's collision, the `collidesWith` array only has the first element itself, so the `getFirstCollision` would return `false`.

So I think the key point is `getFirstCollision` method. If we want same display in two modes ,we should keep the same sequence of `getFirstCollision` 's `layout` param.

I  will create a PR to fix this.

fbambi:
fix #789

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -271,8 +271,13 @@ export function correctBounds(
   bounds: { cols: number }
 ): Layout {
   const collidesWith = getStatics(layout);
-  for (let i = 0, len = layout.length; i < len; i++) {
-    const l = layout[i];
+  const noCollidesWith = getNoStatics(layout);
+
+  // give static priority
+  const orderedLayout = [...collidesWith, ...noCollidesWith]
+
+  for (let i = 0, len = orderedLayout.length; i < len; i++) {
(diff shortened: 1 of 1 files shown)


lorenamelor:
I have the same error heror

Issue #820, opened 2018-05-28
Toolbox demo missing from https://strml.github.io/react-grid-layout/examples/


leothorp:
**expected behavior:**  https://strml.github.io/react-grid-layout/examples/14-toolbox.html leads to a demo of the "toolbox" example.

**current behavior:**  https://strml.github.io/react-grid-layout/examples/14-toolbox.html results in a 404

PR #821, opened 2018-05-28
add missing demo links to README.md

leothorp:
Hi there, noticed that the last three demos are not currently linked in the README; this adds links for them.  NOTE: this isn't ready to merge as is, because the toolbox demo appears to be not hosted on the github site currently. What would be required to fix that? (https://strml.github.io/react-grid-layout/examples/14-toolbox.html results in a 404, issue #820)


STRML:
Thanks. I've updated the examples, so 14 will now resolve. Appreciate it.

PR #826, opened 2018-06-04
Don't collide statics with non-statics

### LINKED ISSUE #825: Static items change place
Generate a layout with static items overlapping non-statics.

Expected behaviour:

Static item should stay where it is.

Observed behaviour:

Static item changes place.

anacierdem:
Otherwise static elements change position due to non-statics.

correctBounds is called just before a compaction pass, so this should not cause other issues.

Fixes #825


anacierdem:
Latest change covers all scenarios

PR #829, opened 2018-06-13
Ensure no negative positions when compacting

### LINKED ISSUE #535: Negative Positioning When Using Non-Integer Grid Units
- [x] Bug (or documentation error)
- [ ] Feature Request
- [ ] Question

There's a _potential_ bug or documentation error with the positioning of the grid items.

According to the Grid Layout Props, layout is an array of objects with this format:
```
// Layout is an array of object with the format:
// {x: number, y: number, w: number, h: number}
```

Seeing that there's nowhere in the usage docs that specifically mention to use integer values, I started using non-integer values with positioning and heights.

The bug I ran into happens when the y value is between 0 and 1. In the `compactItem` function inside `utils.js` we can see the following code: 

```
// Move the element up as far as it can go without colliding.
while (l.y > 0 && !getFirstCollision(compareWith, l)) {
    l.y--;
}
```

This can actually cause the y value to dip below 0 and cause the grid item to break out of the grid layout area.

A possible solution I see here is either do a `l.y = Math.max(l.y, 0)` check after this line or update the usage docs to specifically mention using integer values for grid layout.

Thanks for your time and making this layout system!

DonnyLi:
Ensure no negative positions in util.js compactItem function.

Fixes #535.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -257,6 +257,11 @@ export function compactItem(
       l.y++;
     }
   }
+  
+  // Ensure that there are no negative positions
+  l.y = Math.max(l.y, 0);
+  l.x = Math.max(l.x, 0);
+  
   return l;
 }
 


DonnyLi:
I don't actually use this anymore, but I'd figured I'd try and close the issue that I brought up a year ago.

evan-goode:
Patch works great, fixed an issue that had been bugging us bad. @STRML should be an easy merge 😄

kirandash:
Can this fix be merged. Since the issue #535 still appears.

Issue #626, opened 2017-09-13
[Feature Request] Ability to disable grid editing without removing Resizeable and Draggable wrappers


RyGuyM:
Please mark the type of this issue:

- [ ] Bug
- [*] Feature Request
- [ ] Question

I wanted to get feedback on a feature request before creating PR. 

Ability to disable `draggable` and `resizeable` without removing the Draggable and Resizeable wrappers. This prop would act as a passthrough to those components and their disabled prop (as far as I understand react-resizeable would need a disabled prop added as well).

The use case here is we have some expensive components inside our react-grid-layout. At certain times in our application we want to prevent resizing and dragging. Our first pass we just set `isResizable` and `isDraggable` to false, this caused all of the expensive components to unmount then remount causing a significant performance hit to our UI. We can work around this in the application but it seemed like it could be useful to have it built into react-grid-layout.

bolte-17:
I just stumbled upon this same problem. How are you working around this?

RyGuyM:
First I added css `pointer-events: none` for disabling mouse events on the corresponding class `.react-resizable-handle` and a custom class for the drag handle. On the JS side the same boolean flag that I used to enable the css class to prevent dragging/resizing from propagating up from the event handlers.

pinkynrg:
This improvement would be super nice!

eric-burel:
Hello, this may be considered an issue, not simply a feature request. Without speaking of expensive components, remounting a component will make it lose its state. This prevent using ReactGridLayout as the core layout of the application, since a lot of components may not like mount/unmount.

yuzhakovvv:
+1

larrydahooster:
@eric-burel I don't think there are ways to work around cause it would be very hard to detect the same component in the dom if the path changes on a parent node.

This is a common pitfall. But I would also like to see the feature! +1

mistercrunch:
+1 , I was working on adding a "editable mode" toggle button in my app and my stateful components are getting unmounted. For now I'm working around the issue by hiding my drag handle and will disable click events around resize.

Issue #756, opened 2018-03-15
Is there a way to toggle draggable status without unmout/mount element


m7luffy:
Thanks for submitting an issue to RGL!

Please mark the type of this issue:

- [ ] Bug
- [ ] Feature Request
- [ √] Question

If you have a question or bug report, please use the [WebpackBin Template](https://www.webpackbin.com/bins/-Kvr2qCxorvGMgVMxkmI)
to demonstrate. It is much easier for us to help you if you do.

I need a function to toggle draggable status. I do this by changing 'static' or 'isDraggable' property in layout. But changing these two properties will lead element unmounted. Mount a new element take a while in my app because it require data from network. Is there any way to change the status without re-mount the element?

luminaxster:
**TL;DR:** Try providing your own  ```draggableHandle=".className"```:

Example:

``` <Responsive draggableHandle=".draggable" > 
<div key="x" >
            <div className={isDraggable?'draggable': 'disabled'}> ...</div>
</div>
...
</Responsive>
```

m7luffy:
Thanks @luminaxster , this way does work.

trydzynski-zen:
I'm impacted by this as well. When I toggle isDraggable and isResizable, all widgets are recreated.
My widgets (the draggable component) cache their queries, so remounting them does not cause network load, however they blink and/or animate which looks poorly.

Could this be fixed in the library, e.g. by calling functions, instead of adding Resizable/Draggable mixins via wrapping in new components? That should be enough to unlock react's reuse of components (I have verified the cause, but not the fix).

Issue #721, opened 2018-02-07
GridItem Destroys the Wrapped Child Element


ZheyangSong:
Thanks for submitting an issue to RGL!

Please mark the type of this issue:

- [X] Bug
- [ ] Feature Request
- [X] Question

If you have a question or bug report, please use the [WebpackBin Template](https://www.webpackbin.com/bins/-Kvr2qCxorvGMgVMxkmI)
to demonstrate. It is much easier for us to help you if you do.

**NOTE: I can't get the bin working. The app constantly reports that some error occurs in the template... I'd like to create a simple demo if the error can be fixed...**

For now, I'll describe the issue, expectations and possible solution with code snippets and references...

The code this issue discusses is [here](https://github.com/STRML/react-grid-layout/blob/master/lib/GridItem.jsx#L471-L491).

### Summary
In my case, I use this library to create an editable dashboard (refer to sample code blow). The UI provides a button toggling the editable state of the dashboard via setting `isDraggable` and `isResizable` props of `ReactGridLayout`...But, due to the aforementioned implementation, every time the values of these two props change, the custom element/component wrapped in `GridItem` will be recreated by cloning. Hence, the original wrapped element is destroyed and its internal state is lost...

This seems to imply that every custom item/component passed into the grid layout has to be stateless. If this is true, its not very feasible, since it's not always true for a custom component to be stateless...

```jsx
// dashboard code
// A modification borrowed from https://github.com/STRML/react-grid-layout/blob/master/test/examples/1-basic.jsx

class Dashboard extends React.PureComponent {
  static defaultProps = {
    items: 20,
    rowHeight: 30,
    cols: 12
  }

  constructor(props) {
    super(props);

    const layout = this.generateLayout();

    this.state = {
      editable: false,
      layout
    };
  }

  toggleEditMode = () => {
    this.setState(({editable}) => {
      editable: !editable
    });
  }

  generateLayout() {
    const p = this.props;
    return _.map(new Array(p.items), function(item, i) {
      const y = _.result(p, "y") || Math.ceil(Math.random() * 4) + 1;
      return {
        x: (i * 2) % 12,
        y: Math.floor(i / 6) * y,
        w: 2,
        h: y,
        i: i.toString()
      };
    });
  }

  generateDOM() {
    return _.map(_.range(this.props.items), (i) => (<CustomItem key={i} />));
  }

  render() {
    const {
      editable,
      layout
    } = this.state;

    return (
      <div className="dashboard">
[...]

Issue #379, opened 2016-10-20
Changing isDraggable and isResizable reloads the children


karteek-gamooga:
When the isDraggable or isResizable is changed in responsive grid layout it recreates the children every time. This is because the hierarchy is changed as react-draggable/react-resizable is inserted in the middle and react reloads all the child components. 

This issue can be fixed by always using the react-draggable/react-resizable components and passing isDraggable/isResizable over to the disabled prop of react-draggable/react-resizable. Would be happy to assist in the change.

And thanks for the wonderful library 👍

daumann:
see  #371

daumann:
Similar to that, if you remove one element , it recreates all elements in the set. @STRML do you have any idea how to circumvent this like the "hack" you mentioned in #371 ?

STRML:
Yes, this is an unfortunate issue in the current architecture. I plan to fix this in #346.

PR #892, opened 2018-12-19
fix: dynamic isDraggable and isResizable remount

### LINKED ISSUE #240: dynamic isDraggable isResizable
Hi,
I'm currently dynamically setting isDraggable and isResizable based upon whether you go into an edit mode for our dashboard grid layout.
This works great ... 
Although, whenever one of these changes from false-true and then back again the GridItems are unmounted and remounted.
This means that my statefull components inside a GridItem get unmounted and then remounted .... amounting to their state being blown away.
I wouldn't normally use statefull components but they include c3 graphs which have their own lifecycle outside of react, etc ...
So, is there a way of stopping this behaviour ?
If I change any properties of the child components through a redux connect they don't get unmounted.
Any help would be much appreciated.
/Stephen.

lmachens:
Fixes #240 
Fixes #892

Issue:
Childrens are remounted when isDraggable or isResizable is changed.

Solution:
The `Resizable` and `DraggableCore` parents are always mounted to prevent remount of children.
Use `disabled` to prevent calling drag handlers and hide the resizable handle with css.


lmachens:
@STRML any update?

kravets-levko:
This PR will also fix #626

artemsaveliev:
Made another way to do it #937 
Not speaking for maintainer, but I am not sure switching to always wrapping with DragggableCore is good. In my version you can still remove DraggableCore by setting isDraggable off.

nevalera:
Any updates on this or #937 ?

The fix for the remounting issue is required for us to be able to use this library.

agg23:
Another ping for this PR to get looked at. This is a pretty major blocker for many scenarios.

STRML:
What performance impact will this have? I am also loathe to always wrap elements when not needed. I would prefer the user to manually hide the handles if they have the desire to make draggability/resizability dynamic.

asaveliev:
> I would prefer the user to manually hide the handles if they have the desire to make draggability/resizability dynamic

Currently entering/exiting editing triggers complete reload of all widgets - depending on scenario, this may not be a great experience. Specifically for us it is expensive as widgets have expensive rendering paths. 

I am not sure there's a way to make draggability dynamic with current version - other than switching each widget to static?

n1ghtmare:
> What performance impact will this have? I am also loathe to always wrap elements when not needed. I would prefer the user to manually hide the handles if they have the desire to make draggability/resizability dynamic.

@STRML As far as performance impact, what are your concerns here? Otherwise, I actually tend to agree with you on the second part (in the sense that, we shouldn't clutter the code-base with various features that can be achieved with some code on the user's side). However in this case I think there is a good use case as others stated.

STRML:
This can be done for resizability per item by hiding the handle via CSS, and for grid items by returning `false` from the grid's `onDragStart(i: string, x: number, y: number, { e, node }: GridDragEvent)` callback.

agg23:
While that is good to know, I don't really consider those solutions to be much more than hacks. In general, I expect React components to dynamically respond to configuration updates with only minor rerendering; that is one of the core constructs of React. The fact that there are `isDraggable` and `isResizable` props that require such hacks if you want them to respond in the "expected" way indicates to me that something more should be done in this area.

mitulmk:
Nudge...

mitulmk:
Great to see progress... What is the "Labeler / label (pull_request)" check that's failing?

agg23:
@STRML, are you strongly opposed to this change? As it stands, is there any likelihood that this gets approved, or should this PR be closed?

JohnDDuncanIII:
> This can be done for resizability per item by hiding the handle via CSS, and for grid items by returning false from the grid's onDragStart(layout, oldItem, newItem, placeholder, e, ?node) callback.

The above worked almost perfectly, but since the `GridItem.jsx` `onDragStart` computes and sets the `dragging` state value before it invokes `this.props.onDragStart && _this.props.onDragStart`, this meant that if I clicked on any disabled dashboard widget when not in edit mode, then `react-grid-layout` would still append the `react-draggable-dragging` `className` to the widget.

This caused a particularly nasty issue where any widget that had been clicked before we entered edit mode would get "stuck" in its "dragging" position until you dragged it away and then back in. If you tried to drag a non-stuck widget over the stuck one, they would just overlap with each other.

Screen Shot 2020-02-10 at 3 46 20 PM

To fix this, I took the aforementioned advice from [STRML](https://github.com/STRML/react-grid-layout/pull/892#issuecomment-532014140) and also utilized some techniques outlined in all of the related issues to reach what I would consider to be the current "canonical" solution to the unmount/remount bug that occurs when you change `isDraggable`/`isResizable`

https://github.com/STRML/react-grid-layout/issues/89
https://github.com/STRML/react-grid-layout/issues/240
https://github.com/STRML/react-grid-layout/issues/362
https://github.com/STRML/react-grid-layout/issues/371
https://github.com/STRML/react-grid-layout/issues/379
https://github.com/STRML/react-grid-layout/issues/626
https://github.com/STRML/react-grid-layout/issues/721
https://github.com/STRML/react-grid-layout/issues/756
https://github.com/STRML/react-grid-layout/issues/1067

```jsx
<StyledReactGridLayout
    ...
    // this completely disables the drag functionality for a GridItem
    draggableHandle={!this.props.editing ? ".not-draggable" : null}
    // this is our internal redux store value that tracks whether we have clicked the edit button
    editing={this.props.editing}
    // these should always be set to true to prevent remounts
    isDraggable
    isResizable
    // if we are editing, we want the default onDragStart functionality
    // if we are not editing, then we return false to disable the internal GridItem onDrag
    onDragStart={() => this.props.editing}
    ...
>
    {...}
</StyledReactGridLayout>
```

```javascript
import styled, { css } from "styled-components"
import RGL, { WidthProvider as provideWidth } from "react-grid-layout"
const ReactGridLayout = provideWidth(RGL)

export const StyledReactGridLayout = styled(ReactGridLayout)`
    position: relative;
    transition: height 100ms ease;

    .react-grid-item {
        box-sizing: border-box;
        transition: all 100ms ease;
        transition-property: left, top;

        .react-resizable-handle {
            display: none;
        }

        ${props => (props.editing) && css`
            .react-resizable-handle {
                display: initial;
            }

            &:hover {
                cursor: pointer;
            }

            &:active {
                background: #dadada;
                cursor: move;
            }
        `}
    }
`
```

STRML:
Alright. I think yall have convinced me that this is worth the rather small bit of overhead that is added in cases where you are not using draggability/resizability. I would rather that this component be easy to use. Merging. I'll cut a release in the next few days when I have time to prepare the changelog and do extra regression testing.

STRML:
Will be released in a few minutes in `0.18.0`.

PR #905, opened 2019-01-30
Add graphext to projects using react-grid-layout list

NachoJusticia:
Hi, I am a developer at [Graphext](https://graphext.com/).
Since a few months ago we are using this library in one of the core parts of our product, and it would be a pleasure for us to appear in the README list as company using this library 👍 .
We are very proud of being using `react-grid-layout` as we have benefited a lot from this.

Thank you very much for creating and maintaining this project.


STRML:
Thanks!

Issue #918, opened 2019-03-03
GridItem calcXY calculation is wrong


xiaobao66:
Hi, thank you very much for this awesome grid system.
When the margin property and containerPadding property is not the same, dragging placeholder is far behind the dragging gridItem.
You can see the demo here: [https://codesandbox.io/embed/m94wn5k3y?autoresize=1&fontsize=14&hidenavigation=1&view=preview](https://codesandbox.io/embed/m94wn5k3y?autoresize=1&fontsize=14&hidenavigation=1&view=preview)
The cause of the problem is GridItem's calcXY function.The calculation of left and top value has a little wrong.It should be 

> left = colWidth * x + margin * x + **containerPadding**

 rather than

> left = colWidth * x + margin * x + **margin**

I have solved the problem and create a pull request. By any chance, could you help check the pr

STRML:
Thanks for this. I'll find some time soon to check on it as it appears you are right.

PR #919, opened 2019-03-03
fix GridItem calcXY error calculation

xiaobao66:
Fixed: #918 

Issue:
GridItem calcXY calculation is wrong.

Solution:
Change the calculation of the left and top value. The new calculation is `left = colWidth * x + margin * x + containerPadding`

### DIFF
--- a/lib/GridItem.jsx
+++ b/lib/GridItem.jsx
@@ -224,18 +224,25 @@ export default class GridItem extends React.Component<Props, State> {
    * @return {Object} x and y in grid units.
    */
   calcXY(top: number, left: number): { x: number, y: number } {
-    const { margin, cols, rowHeight, w, h, maxRows } = this.props;
+    const {
+      margin,
+      cols,
+      rowHeight,
+      w,
+      h,
+      maxRows,
+      containerPadding
+    } = this.props;
     const colWidth = this.calcColWidth();
 
-    // left = colWidth * x + margin * (x + 1)
-    // l = cx + m(x+1)
-    // l = cx + mx + m
-    // l - m = cx + mx
-    // l - m = x(c + m)
-    // (l - m) / (c + m) = x
+    // left = colWidth * x + margin * x + containerPadding
+    // l = cx + mx + p
+    // l - p = cx + mx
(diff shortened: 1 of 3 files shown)


STRML:
Indeed #918 is a real bug. If we can get codesandbox working on PRs it would be great to try this out, otherwise we can pull & compare.

STRML:
Verified this fixes the issue locally with the example grid and:

```
          containerPadding={[0, 0]}
          margin={[100, 10]}
```

However we need a regression test.

xiaobao66:
@STRML Sorry, I don't know how to write a regression test for this fix. Do you mean add a example in the test directory?

STRML:
Yes, something in `test/spec` that fails under the old code but succeeds under the new.

xiaobao66:
> Yes, something in `test/spec` that fails under the old code but succeeds under the new.

Sorry, I don't know what fails under the old code. I didn't change `test/spec` and my pr has passed the test.

STRML:
The intention here is to write *a new test* that ensures that your fix doesn't get removed in the future and to ensure that the new behavior we've just identified as "good" stays good.

xiaobao66:
@STRML  I find `test/spec` only test function, but GridItem.jsx is a react component. This is my first time to commit pr, so I don't know how to add a test for react component. Could you give an example?

STRML:
Yeah, we do it in https://github.com/mzabriskie/react-draggable/blob/master/specs/draggable.spec.jsx using react-dom/test-utils & react-test-renderer. [Enzyme](https://airbnb.io/enzyme/) is also an option.

xiaobao66:
Ok, I will look at it first and then spend time adding the test

xiaobao66:
@STRML I have added a test case for the fix.But it show `Run actions/labeler@v2 failed`, 

> ##[error]HttpError: Resource not accessible by integration
> ##[error]Resource not accessible by integration
> ##[error]Node run failed with exit code 1

what does it mean?

MaxMusing:
I put up an updated PR (#2059) that fixes this and includes updated tests.

STRML:

PR #922, opened 2019-03-20
Get viewport width for breakpoints - optional

krisldz:
Hello!
I need to get the viewport width instead of the component in my app. So I've add some optional props. Readme and example are also updated. Please take a look and decide if you want to have it in the master branch. Would be great if you accept the PR, then I can use your repo in my project instead of using mine :)

Greetings,
Chris.


STRML:

Issue #504, opened 2017-03-03
V-0.14.3: onLayoutChange called when layout didn't change


hepiyellow:
I am experiencing calls to onLayoutChange, when the RGL mount (componentDidMount),
although the layout has not changed.
I saw that in version` 0.13.7` you added the `onLayoutMaybeChanged` method, which is supposed to fix it.
But it seems that it doesn't.
I see that the `loads.isEquals` call returns false, for layout arrays.
It fails on the first itemLayout, where the `oldLayout` has the usual `{i,x,y,w,h}` fields,
and the `newLayout` has also these fields: `{...moved, static, minH,minW,maxH,maxW,isDraggable,isResizable}.

mnemanja:
+1

stephenrs:
This one has also bitten me.

damiangreen:
Yes, this is also a problem for me.

hardvain:
Any updates on this? I am also facing this issue>?

sondresj:
A workaround hack for this bug can be found here:
https://codesandbox.io/s/ol94lx21oy (in the ShowcaseLayout.js file).
This sandbox also reproduces the bug and warns the console each time it occurs. 

The workaround works by checking if the oldItem and newItem in the callbacks of onDragStop and onResizeStop are different (using _.isEqual for deep comparison). 
The _layoutCangeWorkaround property cannot reside in the state, as the onLayoutChange callback is invoked in the same callstack as onDragStop/onResizeStop, so unfortunately we have to mutatete it directly. This doesn't have any consequence to the React rendering of the component however, as we don't need to re-render when _layoutChangeWorkaround is modified.

OrhanOzkercin:
I am facing the same issue, `onLayoutChange` method is triggered when the component is first mount.  Is this problem still continuing or am I missing something?

hepiyellow:
I don’t know. I haven’t used this in years. On 2023. Jan 11., at 2:03, Orhan Özkerçin ***@***.***> wrote:﻿
I am facing the same issue, onLayoutChange method is triggered when the component is first mount.  Is this problem still continuing or am I missing something?

moazzam-ali:
Any updates on this?
I'm facing the same issue.

mohd-khan09:
> onLayoutChange

same here

PR #928, opened 2019-04-08
Invoking onLayoutChange only when item was moved or resized

sondresj:
#504 

When clicking a draggable item or when clicking the resize handle on resizable items the onLayoutChange was invoked, presumably because `isEqual(oldLayout, newLayout)` returns `true`. 
I did not investigate why exactly this happens, but regardless I belive the following solution is better, as it doesn't require a deep comparison of oldLayout and newLayout. 

In the event-handler for onDragStop and onResizeStop I'm making the assumption that if `isEqual(oldDragItem, l)` (ie oldItem is the same as newItem) then the layout has **not** changed, and we don't need to invoke `onLayoutMaybeChanged` method. 

If they are not equal however, then the layout must have changed, and we can invoke `this.props.onLayoutChange` directly with `newLayout`.

### DIFF
--- a/lib/ReactGridLayout.jsx
+++ b/lib/ReactGridLayout.jsx
@@ -412,7 +412,6 @@ export default class ReactGridLayout extends React.Component<Props, State> {
 
     // Set state
     const newLayout = compact(layout, this.compactType(), cols);
-    const { oldLayout } = this.state;
     this.setState({
       activeDrag: null,
       layout: newLayout,
@@ -420,7 +419,9 @@ export default class ReactGridLayout extends React.Component<Props, State> {
       oldLayout: null
     });
 
-    this.onLayoutMaybeChanged(newLayout, oldLayout);
+    if(!isEqual(oldDragItem, l)) {
+      this.props.onLayoutChange(newLayout);
+    }
   }
 
   onLayoutMaybeChanged(newLayout: Layout, oldLayout: ?Layout) {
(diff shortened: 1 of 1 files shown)


sondresj:
Repro of bug with a workaround that proves the concept implemented in the PR: 
https://codesandbox.io/s/ol94lx21oy

STRML:
This *should* be good enough, but it is possible to click and drag an item around in a way that causes items to shift below a static element, then drop the item where it started. This would cause a change in the layout but not trigger this handler.

sondresj:
Right, I noticed that behaviour on [this ](https://strml.github.io/react-grid-layout/examples/11-no-vertical-compact.html) example, where there is no vertical compaction. Is this behaviour expected when `verticalCompact` is `true`? If so, the onLayoutMaybeChanged could be invoked as normal when verticalCompact is false, and my fix could be used when it is true. 

But I do find that behaviour in the example a little strange as well, is there a use-case where it is desirable to displace items by moving another item around?

STRML:
Yes it is expected in both cases. It is desirable to displace items because the item you are dragging needs to fit where you put it. It does that by moving other items out of the way.

sondresj:
I get that when you move one item, you may have to displace others. But without any compaction (verticalCompaction/compactionType), is there really a use-case for displacing items by dragging one item around and placing it back where it was? say you drag it around in circles all over, then drop it back where it was. 
When you have verticalCompaction = true, I don't see how you can move an item around, displacing others while moving, then drop it back where it was, and end up getting a different layout than what you started with.

PR #937, opened 2019-05-13
Added isEditable property to support disabling dragging without rerendering

### LINKED ISSUE #721: GridItem Destroys the Wrapped Child Element

artemsaveliev:
Added isEditable property to support disabling dragging without rerendering

In addition:
* added windows bat file so that make is not needed. 
* modified no-drag example to show how the new property can be used
* fixed example template javascript source path

Fixes #721

### DIFF
--- a/build.bat
+++ b/build.bat
@@ -1,0 +1,11 @@
+@ECHO OFF
+IF "%1%"=="dev" (
+    node .\node_modules\webpack-dev-server\bin\webpack-dev-server --config webpack-dev-server.config.js --hot --progress --colors --port 4002 --open --content-base .
+) ELSE IF "%1%" == "build-example" (
+    node .\node_modules\webpack\bin\webpack --config webpack-examples.config.js
+    node ./examples/generate.js
+) ELSE (
+    rmdir /s /q .\build
+    node .\node_modules\babel-cli\bin\babel --stage 0 --out-dir .\build .\lib
+    node .\node_modules\webpack\bin\webpack
+)
--- a/css/styles.css
+++ b/css/styles.css
@@ -32,6 +32,10 @@
   user-select: none;
 }
 
+.react-grid-item.react-resizable-disabled > .react-resizable-handle {
+  visibility: hidden;
+}
+
 .react-grid-item > .react-resizable-handle {
   position: absolute;
   width: 20px;
--- a/examples/0-showcase.html
+++ b/examples/0-showcase.html
@@ -5,7 +5,7 @@
 <html>
   <head>
     <meta charset="UTF-8">
-    <script src="http://localhost:4002/dist/commons.js"></script>
+    <script src="http://localhost:4002/dist/commons.bundle.js"></script>
     <script src="http://localhost:4002/dist/0-showcase.bundle.js"></script>
     <title>RGL Example 0 - Showcase</title>
   </head>
--- a/examples/1-basic.html
+++ b/examples/1-basic.html
@@ -5,7 +5,7 @@
 <html>
   <head>
     <meta charset="UTF-8">
-    <script src="http://localhost:4002/dist/commons.js"></script>
+    <script src="http://localhost:4002/dist/commons.bundle.js"></script>
     <script src="http://localhost:4002/dist/1-basic.bundle.js"></script>
     <title>RGL Example 1 - Basic</title>
   </head>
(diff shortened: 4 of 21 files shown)


STRML:

Issue #961, opened 2019-07-03
onWindowResize not providing correct width when wrapped with resizing container


suhailnaw:
**Type**:
Bug

**Issue**:
In my use case, `ResponsiveReactGridLayout` is nested under a container which also resizes, and that causes `WidthProvider` to get the wrong width when the window is maximized.  I set `measureBeforeMount` to `true` and that doesn't fix the issue.  I get a bug after:
1. resize the window to the smallest breakpoint
2. click maximize on the browser window
3. elements will not correctly resize
4. the width in state is incorrect. if you slightly move the divider between the code and the grid, it will correctly resize, see video.

**Link to sandbox replicating issue**:
https://codesandbox.io/s/broken-waterfall-rmsee
I added a `<div>` container which also resizes in the index.js file with `id="Resizing-Container"`.  This Resizing-Container div has inline style that gives it `width`.  The `width` comes from listening to another `resize` event listener that’s a parent of `ResponsiveReactGridLayout`.  `measureBeforeMount` is set to `true`.

**Video**:
![rgl](https://user-images.githubusercontent.com/26286035/60555062-986e6800-9cef-11e9-9c5a-4449f4a9a086.gif)

**Proposed Fix**:
Change `WidthProvider`’s `onWindowResize` method to wrap the `setState` in a `setTimeout` with a timeout of `0`.  This fixes the issue.  I will open a PR now.

PR #962, opened 2019-07-03
add setTimeout in onWindowResize for Issue #961

suhailnaw:
#961

### DIFF
--- a/lib/components/WidthProvider.jsx
+++ b/lib/components/WidthProvider.jsx
@@ -60,7 +60,7 @@ export default function WidthProvider<
       // eslint-disable-next-line
       const node = ReactDOM.findDOMNode(this); // Flow casts this to Text | Element
       if (node instanceof HTMLElement)
-        this.setState({ width: node.offsetWidth });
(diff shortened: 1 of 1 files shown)


lib/components/WidthProvider.jsx
```diff
@@ -60,7 +60,7 @@ export default function WidthProvider<
       // eslint-disable-next-line
       const node = ReactDOM.findDOMNode(this); // Flow casts this to Text | Element
       if (node instanceof HTMLElement)
-        this.setState({ width: node.offsetWidth });
+        setTimeout(this.setState({ width: node.offsetWidth }), 0);
```
daynin: I think making setTimeout on each event loop tick is not a good solution. Unfortunately, this change will negatively affect performance

n1ghtmare: I believe a good rule of thumb is that anytime you use `setTimeout(() => ..., 0)` in a react component, is a sign that the solution needs to be re-engineered, or at the very least a significant code smell.

STRML:

Issue #784, opened 2018-04-02
Several bugs with horizontal compactType


jordanbb:
- [x] Bug
- [x] Feature Request
- [ ] Question

I'm loving this library so far but I've noticed a few bugs with horizontal compactType. These bugs seem  to be unique to horizontal mode. I've outlined what I've run into below.

## 1. Collision between different sized draggables along the same row.

![rgl-different-size-reorder-bug](https://user-images.githubusercontent.com/13209073/38209261-053940f4-3671-11e8-8faa-1037af2845a2.gif)

This appears to only happen if your dragging a smaller item near a larger item. In the gif the expected behavior is that the item being dragged should have swapped places with the larger item.

## 2. Dragging can cause the layout to re order over and over rapidly 
![rgl-horizontal-crazy-re-ordering](https://user-images.githubusercontent.com/13209073/38209928-149d3bd4-3673-11e8-851f-eeab32bcbe0c.gif)

This appears to only happen if you are dragging an item from the left most column.

## 3. Collision issues with the items in the left most column when draggin up or down.
![rgl-horizontal-moving-between-rows](https://user-images.githubusercontent.com/13209073/38210289-3f845f34-3674-11e8-88b8-0cea8528c0bc.gif)

When dragging an item in the left most column up or down into a new row, I would think that it should displace items to the right to make room, instead of swapping positions. The collision issue before the items swap appears to only be if your dragging an item into the bottom most row.

## 4. Prevent layout from changing unintentionally when dragging
Horizontal mode should probably create a column (similar to vertical creating a row) when dragging around so as not to force items to wrap to a new row. This behavior changes the layout as the user drags and can't be corrected by simply moving the item back to where you started like you can with vertical mode. This isn't really a bug but a suggested feature, but one that would make horizontal mode feel better in my opinion.

I may have some time to contribute and submit a PR for these fixes. I'm curious if anyone else may be working on any of these fixes though?

STRML:
These issues are odd. I'm unaware of most of them. If you can write a failing test it would be very helpful and we can then collaborate on a fix in a PR.

jordanbb:
Id love to help anyway I can. By failing tests do you mean a live example of these issues hosted on something like codesandbox.io?

rezaabedian:
I have issue 2) happening in a project of mine. @jordanbb did you figure anything out?

rezaabedian:
[...]

PR #967, opened 2019-07-16
Fix horizontal compact issue for Issue #784

pjlake98:
#784 

This fixes the issue where dragging could cause the layout to reorder over and over rapidly.


littletower:
this does not fix #784 at all.

STRML:

PR #975, opened 2019-08-06
changed readme accordingly to responsive mechanism

vichugunov:
Readme is a bit confusing, because accordingly to [example](https://github.com/STRML/react-grid-layout/blob/3352c84515e587c2bb7ec7935e24be5705d33704/test/examples/8-localstorage-responsive.jsx#L49) it's possible to use `data-grid` props to set default mappings for individual items


STRML:
Thanks!

Issue #586, opened 2017-07-05
How to drag and drop between two grids or from outside the grid (dashboard)


razvantomegea:
Thanks for submitting an issue to RGL!

Please mark the type of this issue:

- [ ] Bug
- [ ] Feature Request
- [x] Question

If you have a question or bug report, please use the [WebpackBin Template](http://www.webpackbin.com/EJF48h_rz)
to demonstrate. It is much easier for us to help you if you do.

razvantomegea:
It seems this question was already mentioned
https://github.com/STRML/react-grid-layout/issues/459

daynin:
@razvantomegea we added an ability to drag some elements from outside of the grid

PR #980, opened 2019-08-15
feat(grid): add an ability to drop elements from outside of a grid

daynin:
- add isDroppable flag for grids
- add onDrop callback for getting an element position after dropping
- add documentation
- add an example

Relates to: https://github.com/STRML/react-grid-layout/issues/586


daynin:
@n1ghtmare @josesaranda @mankinchi hi! You said here https://github.com/STRML/react-grid-layout/issues/985 you want to help the project. So could you review this PR pls? And you can mention me in your PRs, I'll review them too

lib/GridItem.jsx
```diff
@@ -161,6 +169,42 @@ export default class GridItem extends React.Component<Props, State> {
     className: ""
   };
 
+  currentNode: HTMLElement;
+
+  componentWillReceiveProps(nextProps: Props) {
```
n1ghtmare: As you might be aware, `componentWillReceiveProps`  will be unsupported as of React version 17. [Here](https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops) is more info. We should try to stay away from it, and in fact re-write the logic in other places of the project.

lib/ReactGridLayout.jsx
```diff
@@ -621,14 +654,68 @@ export default class ReactGridLayout extends React.Component<Props, State> {
...
     );
   }
 
+  onDragOver = (e: DragOverEvent) => {
+    const { droppingItem } = this.props;
+    const { layout } = this.state;
+    const { layerX, layerY } = e.nativeEvent;
+    const droppingPosition = { x: layerX, y: layerY, e };
+
+    if (!this.state.droppingElem) {
```
n1ghtmare: I don't have a good suggestion here (and sorry this is a bit nit-picky, feel free to ignore it), but is there any way we can rename the `droppingItem` and `droppingElem` so that  the difference is a bit clearer.

test/examples/0-showcase.jsx
```diff
@@ -84,18 +88,23 @@ class ShowcaseLayout extends React.Component {
         <button onClick={this.onCompactTypeChange}>
           Change Compaction Type
         </button>
+        <div className="droppable-element" draggable={true} unselectable="on">
```
n1ghtmare: This is cool, it will be even better if this awesome feature has its own separate page for a demo I think.

lib/GridItem.jsx
daynin: @n1ghtmare I have a solution in my fork. I think it'll be better to replace all componentWillReceiveProps methods in a separate PR

lib/ReactGridLayout.jsx
daynin: @n1ghtmare done

test/examples/0-showcase.jsx
daynin: @n1ghtmare There are no links between documentation pages, so if I move this example into separate page, users won't be able to see it. I think we should update documentation system. I can replace current docs with https://docusaurus.io/ (or something like that)

What do you think?

n1ghtmare: Sure, in the long run, having a better docs page would be great, I fully agree. However for now, why not just add it to the demo selection in the README as number 16 (or whatever the next number was) and have it be on its own (same like the rest, with "view source" and all).

daynin: @n1ghtmare done

mankinchi:
@n1ghtmare sorry I'm new to review code on github. Can you show me how?

n1ghtmare:
@mankinchi Sure thing. First look at the top of this pull request. You will see several tabs, the first one being "Conversation" which is the one we're on now. The next one is "Commits", click on it and select the commit that you want to review (in this case it's just one). You'll now see the commit and the changes. Go through them (through the actual code) and leave a comment on any line that you feel needs a comment/clarification/suggestion or whatever. You might not have any comments of course. Once you're done, in the upper right corner you'll see "Review Changes". Click it and select the option you need. Approve accordingly.

I hope this helps. Let me know if you need more clarifications.

mankinchi:
@n1ghtmare thanks. I am able to fetch and checkout this PR branch but couldn't get the server to run. I'm using `npm run dev` but it only return the showcase page. How can I check the example page? If I'm running `npm run view-example`, I got problem with `/dist/xxx` file not found. Can you show me how? I think we also need to provide a guide on how to start run/build the project locally

mankinchi:
Ok figured out the reason why. I think this is out of the scope of this task but not sure whether we can address it here. The build-example is pumping out `common.bundle.js` but all of the example is looking for `common.js`, which causes the example to fail. Changing in `template.ejs` make it works

daynin:
@mankinchi 
> I mean it make senses to do it, but I think we would want to have a separate callback for it.
Thank you for your comment! But I think it's ok to use the same method for doing the same thing. This code uses onDragStart for dragging simulation, that's why we need to reuse it here

daynin:
@STRML check it pls. If it's ok, I'll merge it

STRML:
See comments above - forgot to hit the "submit review" button :(

mankinchi:
will this be pushed to npm soon or it requires the fixes for @STRML's comment first?

lib/GridItem.jsx
```diff
@@ -161,6 +169,42 @@ export default class GridItem extends React.Component<Props, State> {
...
+    const { droppingPosition } = nextProps;
+    const { dragging } = this.state;
+
+    if (!droppingPosition || !this.props.droppingPosition) {
+      return;
+    }
+
+    if (!this.currentNode) {
+      // eslint-disable-next-line react/no-find-dom-node
+      this.currentNode = ((ReactDOM.findDOMNode(this): any): HTMLElement);
```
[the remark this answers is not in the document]
daynin: @STRML no, I think it's impossible

lib/GridItem.jsx
```diff
@@ -161,6 +169,42 @@ export default class GridItem extends React.Component<Props, State> {
...
+    }
+
+    if (!dragging) {
```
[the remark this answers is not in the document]
daynin: @STRML if you don't mind I'll refactor it here https://github.com/STRML/react-grid-layout/pull/990

test/examples/16-drag-from-outside.jsx
```diff
@@ -0,0 +1,134 @@
+import React from "react";
```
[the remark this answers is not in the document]
daynin: @STRML fixed https://github.com/STRML/react-grid-layout/pull/993

violabg:
Hi I think you should update @types/react-grid-layout too, since is at 0.16.7 version and doesn't have any drop related typings.
Thank for this feature.

STRML:
I don't maintain that package.

mikefordweb:
> * add isDroppable flag for grids
> * add onDrop callback for getting an element position after dropping
> * add documentation
> * add an example
> 
> 
> 
> Relates to: #586



> * add isDroppable flag for grids
> * add onDrop callback for getting an element position after dropping
> * add documentation
> * add an example
> 
> 
> 
> Relates to: #586

How can I do this??

mathiastm:
Hi, 
Did someone can tell me where I can find this exemple please ? 
I really need that...

kryptoscan:
Anyway you can combine widgets into one widget with tabs?

detautama:
@mikefordweb 
@mathiastm 
This is the example I’ve tried to create.
https://codesandbox.io/p/devbox/react-dnd-tailwind-yngc8r

Demo [GIF] bellow (take time to loading)
react-grid-layout-widget-dnd-resizeable-example

PR #981, opened 2019-08-16
Ability to store data on layout items

oguzgelal:
I had a use case where I needed to attach certain data to each grid item. The issue is that custom data points disappear after every layout change (happens in `utils.js/cloneLayoutItem`). I added a special data point, `data`, which will be preserved.


daynin:
@oguzgelal thank you for your contribution! But can we solve this problem by storing data in some object like this:

```javascript
const store = {
  'id1': { /* some data here */ },
  'id2: { /* some data here */ },
}
```
and getting the data by GridItem id?

PR #982, opened 2019-08-16
Rd/refactoring

RuslanDenha:

example or test case.


STRML:
While this is a nice refactor to functional components and hooks, does this actually improve performance or extensibility?

Issue #988, opened 2019-09-05
Console warnings using react 16.9


gknapp:
My app console warnings:
![image](https://user-images.githubusercontent.com/143027/64346913-ea32b780-cfea-11e9-9c5d-28ff0c93c63b.png)

From [code sandbox example](https://codesandbox.io/embed/headless-snow-nq18m) (with upgraded react 16.9.0 deps):
![image](https://user-images.githubusercontent.com/143027/64347130-4dbce500-cfeb-11e9-9a1b-e1abf9839a6d.png)

Using react-grid-layout 0.16.6

daynin:
I'm working on it. I'll try to fix it as soon as possible

n1ghtmare:
@daynin I briefly looked into it earlier today when I got some free time, but from the looks of it I had to change quite too many things in the internals to make it work with `getDerivedStateFromProps` and `componentDidUpdate`, perhaps you're on a better path than me. Let me know if you need any help with it.

daynin:
> but from the looks of it I had to change quite too many things in the internals to make it work with getDerivedStateFromProps and componentDidUpdate

@n1ghtmare I think I did it the same way, check it out pls

daynin:
I fixed it here #990

PR #990, opened 2019-09-09
fix: replace componentWillReceiveProps by getDerivedStateFromProps and componentDidUpdate

daynin:
relates to: https://github.com/STRML/react-grid-layout/issues/988

### DIFF
--- a/lib/GridItem.jsx
+++ b/lib/GridItem.jsx
@@ -171,11 +171,17 @@ export default class GridItem extends React.Component<Props, State> {
 
   currentNode: HTMLElement;
 
-  componentWillReceiveProps(nextProps: Props) {
-    const { droppingPosition } = nextProps;
+  componentDidUpdate(prevProps: Props) {
+    if (this.props.droppingPosition && prevProps.droppingPosition) {
+      this.moveDroppingItem(prevProps);
+    }
+  }
+
+  moveDroppingItem(prevProps: Props) {
+    const { droppingPosition } = this.props;
     const { dragging } = this.state;
 
-    if (!droppingPosition || !this.props.droppingPosition) {
+    if (!droppingPosition || !prevProps.droppingPosition) {
       return;
     }
 
@@ -184,20 +190,21 @@ export default class GridItem extends React.Component<Props, State> {
       this.currentNode = ((ReactDOM.findDOMNode(this): any): HTMLElement);
     }
 
+    const shouldDrag =
+      (dragging && droppingPosition.x !== prevProps.droppingPosition.x) ||
+      droppingPosition.y !== prevProps.droppingPosition.y;
+
     if (!dragging) {
-      this.onDragHandler("onDragStart")(droppingPosition.e, {
+      this.onDragStart(droppingPosition.e, {
         node: this.currentNode,
         deltaX: droppingPosition.x,
         deltaY: droppingPosition.y
       });
-    } else if (
-      (dragging && droppingPosition.x !== this.props.droppingPosition.x) ||
-      droppingPosition.y !== this.props.droppingPosition.y
-    ) {
+    } else if (shouldDrag) {
       const deltaX = droppingPosition.x - dragging.left;
       const deltaY = droppingPosition.y - dragging.top;
 
-      this.onDragHandler("onDrag")(droppingPosition.e, {
+      this.onDrag(droppingPosition.e, {
         node: this.currentNode,
         deltaX,
         deltaY
@@ -355,9 +362,9 @@ export default class GridItem extends React.Component<Props, State> {
   mixinDraggable(child: ReactElement<any>): ReactElement<any> {
     return (
       <DraggableCore
-        onStart={this.onDragHandler("onDragStart")}
-        onDrag={this.onDragHandler("onDrag")}
-        onStop={this.onDragHandler("onDragStop")}
+        onStart={this.onDragStart}
+        onDrag={this.onDrag}
+        onStop={this.onDragStop}
         handle={this.props.handle}
         cancel={
           ".react-resizable-handle" +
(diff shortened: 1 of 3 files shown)


lib/ReactGridLayout.jsx
```diff
@@ -290,20 +301,20 @@ export default class ReactGridLayout extends React.Component<Props, State> {
     this.onLayoutMaybeChanged(this.state.layout, this.props.layout);
   }
 
-  componentWillReceiveProps(nextProps: Props) {
+  static getDerivedStateFromProps(nextProps: Props, prevState: State) {
```
n1ghtmare: There is a really big performance impact with those changes. I'm getting lag, since when we use `getDerivedStateFromProps` and `componentDidUpdate` it does the equality checks every single time. They're not equivalent to `componentWillReceiveProps` which is only triggered when *props* change (not state). Not sure how we can optimize this.

Also if we go this route, we can do the changes without `getDerivedStateFromProps` and just go with `componentDidUpdate` (with adding a check of what changed and trigger state changes only when needed).

n1ghtmare: I did some changes locally on my branch very similar to yours, but it impacts performance (maybe we can do some measurements and see how bad the impact actually is). Nevertheless, I think for such big changes it will be a good idea to involve @STRML and see what he thinks.

lib/ReactGridLayout.jsx
```diff
@@ -85,6 +89,12 @@ export type Props = {
 };
 // End Types
 
+const compactType = (props: ?Object): CompactType => {
```
STRML: Type here is `Props`, it should not be optional and it should not be `Object`

lib/GridItem.jsx
```diff
@@ -171,11 +171,11 @@ export default class GridItem extends React.Component<Props, State> {
...
   currentNode: HTMLElement;
 
-  componentWillReceiveProps(nextProps: Props) {
-    const { droppingPosition } = nextProps;
+  componentDidUpdate(prevProps: Props) {
+    const { droppingPosition } = this.props;
     const { dragging } = this.state;
 
-    if (!droppingPosition || !this.props.droppingPosition) {
+    if (!droppingPosition || !prevProps.droppingPosition) {
```
STRML: TBH this bailout here makes this code hard to read. #980 should have refactored the droppable functionality into another function so this logic is more readable.

lib/ReactGridLayout.jsx
STRML: Yeah this is problematic, this whole block is going to run many times while dragging an item. We'll need to profile & brainstorm where we can optimize.

daynin:
@n1ghtmare 
> I've left some comments, also, just FYI, I'm getting some errors in the console. Didn't see those before

I fixed it

lib/ReactGridLayout.jsx
daynin: @STRML @n1ghtmare we can try this

> Also if we go this route, we can do the changes without getDerivedStateFromProps and just go with componentDidUpdate

with props mirroring. It can help us to avoid extra comparisons

What do you think?

lib/GridItem.jsx
daynin: @STRML thank you for the comment! I'll fix it right after the performance issue

lib/ReactGridLayout.jsx
STRML: We may want to require the library user to send a new `layout` object (rather than mutating the existing one) or supply something optional like a `layoutDidUpdate` function of arity `(newLayout: Layout, oldLayout: Layout) => boolean` where the default implementation is `(newLayout, oldLayout) => newLayout !== oldLayout`.

This will help us avoid an expensive deep comparison.

Perhaps some of you know a way this is handled in other libraries? I don't want to require the author to mutate `key`, which will unmount all children.

daynin: I think that layout recreating on the user side is not an obvious way to fix it. I mean not obvious for a user

Maybe replacing the code into componentDidMount will be enough to fix performance issues. I'll try to compare cases with componentWillReceiveProps, getDerivedStateFromProps + componentDidUpdate and componentDidUpdate

n1ghtmare: @STRML Just for the sake of brainstorming (and I'm sure I'm missing something), but are we doing this comparison only for when the `compactType` changes or when an item (child) is added/removed?

lib/ReactGridLayout.jsx
```diff
@@ -312,12 +328,27 @@ export default class ReactGridLayout extends React.Component<Props, State> {
...
+        children: nextProps.children,
+        propsLayout: nextProps.layout
+      };
     }
+
+    return null;
+  }
+
+  componentDidUpdate(prevProps: Props, prevState: State) {
+    const newLayout = this.state.layout;
```
n1ghtmare: This still triggers too many times (give it a shot with a `console.log`). Do think we solve it by changing it to the following:

```
    if (!this.state.activeDrag) {
      const newLayout = this.state.layout;
      const oldLayout = prevState.layout;

      this.onLayoutMaybeChanged(newLayout, oldLayout);
    }
```

Similar to what you're doing in the `getDerivedStateFromProps`, or will this break something? I tried it locally and everything seems to be working just fine. However it should improve performance even more. What do you think?

daynin: @n1ghtmare done

n1ghtmare: @daynin From what I see, this change shouldn't break anything. Can you confirm this? Otherwise performance is great. @STRML Do you want to take a quick look at this, when you get some time?

STRML: That's a nice optimization. It shouldn't break anything as the grid should only move in two circumstances:

1. An active drag, or
2. An update to the `layout` property from the parent. 

2) doesn't apply here so we should be okay :+1:

n1ghtmare:
@STRML Can you please change version and push this to npm when you get the chance?

hd4ng:
@daynin @n1ghtmare The `onLayoutChange` is triggered 2 times when dragging an item. 
On version 0.16.6, it works correctly.
Code sandbox: https://codesandbox.io/s/red-waterfall-og2gv
Could you have anything other workarounds for this? Thanks

PR #993, opened 2019-09-11
fix(examples): fix typo 16-drag-from-outside -> 15-drag-from-outside

daynin:


STRML:
`14-toolbox.html` is missing the "next" link to 15

daynin:
@STRML fixed. And I fixed html markup in these files

PR #995, opened 2019-09-15
chore: update FlowType version

daynin:


yarn.lock
```diff
@@ -5124,6 +5124,11 @@ lodash@^4.13.1, lodash@^4.17.10, lodash@^4.17.4, lodash@^4.17.5, lodash@^4.2.0,
   resolved "https://registry.yarnpkg.com/lodash/-/lodash-4.17.11.tgz#b39ea6229ef607ecd89e2c8df12536891cac9b8d"
   integrity sha512-cQKh8igo5QUhZ7lg38DYWAxMvjSAKG0A8wGSVimP07SIUEK2UO+arSRKbRZWtelMtN5V0Hkwh5ryOto/SshYIg==
 
+lodash@^4.17.15:
```
STRML: Should be possible to dedupe this with the previous lodash resolutions

n1ghtmare:
@STRML @daynin I see that in the `react-resizable` repo there is a PR to get rid of `componentWillReceiveProps`, once it's merged and its version is updated, perhaps it will be a good idea to update all our dependencies here (before we push a new version). Let me know I can get right on it.

daynin:
@n1ghtmare 
> once it's merged and its version is updated, perhaps it will be a good idea to update all our dependencies here (before we push a new version)
yes, I think that's a good idea!

I created an issue for it https://github.com/STRML/react-grid-layout/issues/1002

Issue #1004, opened 2019-09-20
drag from outside doesn't work in firefox


daynin:
this example doesn't work https://strml.github.io/react-grid-layout/examples/15-drag-from-outside.html in firefox (version: 68.0.1 (64-bit))

daynin:
This is an old Firefox issue https://bugzilla.mozilla.org/show_bug.cgi?id=568313. We can fix it by adding `ondragstart="event.dataTransfer.setData('text/plain', '')"` in a tag with `draggable` attribute. But it works not properly in our case. I'll try to find out how to fix it

soundyogi:
Hi,
The example works for me but I cant reproduce it in my react app.
draggable='true' is not working.

Is this a react issue?

daynin:
@soundyogi draggable='true' doesn't work in firefox for now. It should work in chrome

daynin:
@soundyogi and you should set `isDroppable` in your ReactGridLayout/ResponsiveReactGridLayout

kenakaiser:
Doesn't work in Chrome for me - The onDrop function does not run - I dropped your source code into my react project and it doesn't work - is there something   missing from this example https://github.com/STRML/react-grid-layout/blob/master/test/examples/15-drag-from-outside.jsx

kenakaiser:
never mind realized not released yet

soundyogi:
I cant get draggable=true to work in any app on any browser. (only the example runs yes,)
I feel like I am loosing my mind I have to setup a codepen tomorrow o.O

soundyogi:
> @soundyogi and you should set `isDroppable` in your ReactGridLayout/ResponsiveReactGridLayout. Without this prop it won't work

I cant even get the element to be draggable at this point. And yes I set up the properties on RGL.
Atm we are using a hackish solution with react-dnd.

kenakaiser:
@soundyogi - they have't released it yet - you would have to use master at this point - not the npm package - also draggable part works on element outside of Responsive - use draggable = {true}. hopefully they will cut a new release version soon

trunks1993:
how to fix it, it dosen’t work in chrome,too

soundyogi:
Ok I removed all react-dnd code from the app, now I can drag elements with draggable=true.

But RGL does not recognize something being dropped.

DO I understand correctly that this feature is NOT in the current npm package 0.16.6 ?

Thanks

trunks1993:
> (I cant get it to work)
[...]

PR #1010, opened 2019-09-26
fix isDroppable in firefox

daynin:
relates to https://github.com/STRML/react-grid-layout/issues/1004

### DIFF
--- a/README.md
+++ b/README.md
@@ -309,10 +309,16 @@ transformScale: ?number = 1,
 // dragged over.
 preventCollision: ?boolean = false;
 
-// If true, droppable elements (with "droppable" attribute) 
+// If true, droppable elements (with `draggable={true}` attribute) 
 // can be dropped on the grid. It triggers "onDrop" callback
 // with position and event object as parameters. 
 // It can be useful for dropping an element in a specific position
+//
+// NOTE: In case of using Firefox you should add
+// `onDragStart={e => e.dataTransfer.setData('text/plain', '')}` attribute
+// along with `draggable={true}` otherwise this feature will work incorrect.
+// onDragStart attribute is required for Firefox for a dragging initialization
+// @see https://bugzilla.mozilla.org/show_bug.cgi?id=568313
 isDroppable: ?boolean = false
 
 //
--- a/lib/ReactGridLayout.jsx
+++ b/lib/ReactGridLayout.jsx
@@ -96,6 +96,9 @@ const compactType = (props: Props): CompactType => {
   return verticalCompact === false ? null : compactType;
 };
 
+const layoutClassName = "react-grid-layout";
+const isFirefox = navigator.userAgent.toLowerCase().includes("firefox");
+
 /**
  * A reactive, fluid grid layout with draggable, resizable components.
  */
@@ -699,6 +702,15 @@ export default class ReactGridLayout extends React.Component<Props, State> {
   }
 
   onDragOver = (e: DragOverEvent) => {
+    // we should ignore events from layout's children in Firefox
+    // to avoid unpredictable jumping of a dropping placeholder
+    if (
+      isFirefox &&
+      !e.nativeEvent.target.className.includes(layoutClassName)
+    ) {
+      return false;
+    }
+
     const { droppingItem } = this.props;
     const { layout } = this.state;
     const { layerX, layerY } = e.nativeEvent;
@@ -754,7 +766,7 @@ export default class ReactGridLayout extends React.Component<Props, State> {
   render() {
     const { className, style, isDroppable } = this.props;
 
-    const mergedClassName = classNames("react-grid-layout", className);
+    const mergedClassName = classNames(layoutClassName, className);
     const mergedStyle = {
       height: this.containerHeight(),
       ...style
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -49,7 +49,10 @@ export type GridResizeEvent = { e: Event, node: HTMLElement, size: Size };
 export type DragOverEvent = MouseEvent & {
   nativeEvent: {
     layerX: number,
-    layerY: number
+    layerY: number,
+    target: {
+      className: String
+    }
   }
 };
 
--- a/test/examples/15-drag-from-outside.jsx
+++ b/test/examples/15-drag-from-outside.jsx
@@ -88,7 +88,16 @@ class ShowcaseLayout extends React.Component {
         <button onClick={this.onCompactTypeChange}>
           Change Compaction Type
         </button>
-        <div className="droppable-element" draggable={true} unselectable="on">
+        <div
+          className="droppable-element"
+          draggable={true}
+          unselectable="on"
+          // this is a hack for firefox
+          // Firefox requires some kind of initialization
+          // which we can do by adding this attribute
+          // @see https://bugzilla.mozilla.org/show_bug.cgi?id=568313
+          onDragStart={e => e.dataTransfer.setData("text/plain", "")}
+        >
           Droppable Element
         </div>
         <ResponsiveReactGridLayout


test/examples/15-drag-from-outside.jsx
```diff
@@ -88,7 +88,13 @@ class ShowcaseLayout extends React.Component {
         <button onClick={this.onCompactTypeChange}>
           Change Compaction Type
         </button>
-        <div className="droppable-element" draggable={true} unselectable="on">
+        <div
+          className="droppable-element"
+          draggable={true}
+          unselectable="on"
+          // this is a hack for firefox
+          onDragStart={e => e.dataTransfer.setData("text/plain", "")}
```
STRML: ?????

lib/utils.js
```diff
@@ -383,9 +388,11 @@ export function moveElement(
   // nearest collision.
   let sorted = sortLayoutItems(layout, compactType);
   const movingUp =
-    compactType === "vertical" && typeof y === "number" ? oldY >= y
-    : compactType === "horizontal" && typeof x === "number" ? oldX >= x
-    : false;
+    compactType === "vertical" && typeof y === "number"
```
STRML: These kinds of changes should be in a lint PR not bundled with other changes. This makes the commit history & review process more confusing.

daynin: @STRML these changes are produced by https://github.com/STRML/react-grid-layout/blob/master/package.json#L16. I think we can't skip them here (or we should change precommit hooks/prettier rules)

test/examples/15-drag-from-outside.jsx
daynin: @STRML I added description about this firefox behaviour

README.md
```diff
@@ -309,10 +309,16 @@ transformScale: ?number = 1,
...
 preventCollision: ?boolean = false;
 
-// If true, droppable elements (with "droppable" attribute) 
+// If true, droppable elements (with `draggable={true}` attribute) 
 // can be dropped on the grid. It triggers "onDrop" callback
 // with position and event object as parameters. 
 // It can be useful for dropping an element in a specific position
+//
+// NOTE: In case of using Firefox you should add
+// `onDragStart={e => e.dataTransfer.setData('text/plain', '')}` attribute
```
STRML: Interesting. Does it actually have to be this bizarre `e.dataTransfer.setData()` bit, or can it just be an empty function?

lib/utils.js
STRML: Okay, I'll run prettier on `master` to eliminate this diff.

STRML:
Could you please rebase on `master`, I've run `prettier` on it.

README.md
n1ghtmare: I agree with @STRML this looks really user-unfriendly. Is there no workaround for this? Sorry I'm on vacation and it's a bit hard to run the project atm.

lib/ReactGridLayout.jsx
```diff
@@ -96,6 +96,9 @@ const compactType = (props: Props): CompactType => {
   return verticalCompact === false ? null : compactType;
 };
 
+const layoutClassName = "react-grid-layout";
+const isFirefox = navigator.userAgent.toLowerCase().includes("firefox");
```
n1ghtmare: We're doing this check anyway, can we perhaps take advantage of it and run `e.dataTransfer.setData('text/plain', '')` for the user within `onDragStart` perhaps?

README.md
daynin: @STRML @n1ghtmare unfortunately, without that bizarre code "draggable" won't work in firefox

lib/ReactGridLayout.jsx
daynin: @n1ghtmare we can't do that because users should place it into their own code (in an element outside of ReactGridLayout)

daynin: It's not about our implementation, it's about "draggable" implementation in firefox

daynin:
@STRML 
> Could you please rebase on master, I've run prettier on it.

done

daynin:
@STRML check it out, pls

STRML:
From your testing, is `e.dataTransfer.setData('text/plain', '')` specifically the code that's needed, or do you just need to access a property on `e`, or does a callback just need to exist?

daynin:
@STRML this code is needed. We need to add this code because of firefox's specificity:

```
Firefox requires 'something' (that we call 'init' here) to be set in dragstart event to initialize the rest of drag events to occur.

This is probably because all of the DOM elements are draggable="true" by default in XUL. (reference: https://bugzilla.mozilla.org/show_bug.cgi?id=646823#c4)

Example:

<div id="something" draggable="true" ondragstart="event.dataTransfer.setData('text/plain', 'node');">Drag me</div>

Chrome doesn't require such 'initialization'.
```
This is from https://bugzilla.mozilla.org/show_bug.cgi?id=568313

And you can find this example of code here https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API/Drag_operations#draggableattribute as well

daynin:
@STRML but this code is external for us. Users need to add this code in their droppable elements for firefox only (this is a common case for draggable API in firefox)

daynin:
I think we should merge this PR and release the next version. @STRML @n1ghtmare what do you think?

STRML:
I understand what you're saying, but what I'm asking is, does it **specifically need to be `e.dataTransfer.setData(...)** or will some other, simpler code do? Otherwise I expect we'll get quite a bit of confusion from users and issues opened about this. It's not at all intuitive.

n1ghtmare:
@STRML @daynin Yeah I was just about to write something similar. On one hand we need to merge this on the other I can foresee a lot of issues opened on this.

daynin:
@STRML 
> it specifically need to be `e.dataTransfer.setData(...)
Yes (as far as I know). And this behaviour is described here https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API/Drag_operations#draggableattribute. It's a known problem in firefox implementation of draggable API

> It's not at all intuitive
I agree, but if you search "draggable in firefox" you'll find many questions about this specificity of firefox. So I think we can consider this behaviour as a known issue

For example. If we know about some strange behaviour of some CSS property we add CSS prefix. So in this case users should use `e.dataTransfer.setData('text/plain', '')` like a CSS property prefix

daynin:
And I added information about this issue in the documentation, so I hope users will find answers there

n1ghtmare:
Maybe we can work on the documentation and refer users to it for this specific edge case. Meanwhile we can keep an eye at the firefox bug tracker for a possible fix.

daynin:
@n1ghtmare

> Maybe we can work on the documentation and refer users to it for this specific edge case.

I've already done it (https://github.com/STRML/react-grid-layout/pull/1010/files#diff-04c6e90faac2675aa89e2176d2eec7d8R317)

> Meanwhile we can keep an eye at the firefox bug tracker for a possible fix.

This issue (https://bugzilla.mozilla.org/show_bug.cgi?id=568313) was created 10 years ago so I think they won't fix it in the nearest future  :)

lib/ReactGridLayout.jsx
```diff
@@ -699,6 +702,14 @@ export default class ReactGridLayout extends React.Component<Props, State> {
   }
 
   onDragOver = (e: DragOverEvent) => {
+    // ignore events from layout's children
+    if (
```
STRML: In my testing, simply adding `event.dataTransfer.setData('text/plain', '')` fixes the bug. Why is this exception needed?

STRML:
Btw, noticed another bug while testing this: if you let go of the droppable element without dropping it, the placeholder still remains:

image

lib/ReactGridLayout.jsx
daynin: in case of firefox dragging fires events with different targets and it makes dragging inaccurate (the placeholder can jump unpredictably without this code)

STRML: Ah I can reproduce that. Thanks. Could you update the comment to indicate this for the future?

STRML: Ah I can reproduce that. Thanks. Could you update the comment to indicate this for the future?

STRML: Ah I can reproduce that. Thanks. Could you update the comment to indicate this for the future?

daynin:
> Btw, noticed another bug while testing this: if you let go of the droppable element at the top of the screen (without dropping it into the layout), the placeholder still remains

@STRML I'll fix it in another PR

STRML:
Ok. Please add some more instructive comments on the event ignoring as indicated above, then this is okay to merge. Thanks for tackling a difficult/weird issue.

daynin:
@STRML 
> Please add some more instructive comments on the event ignoring as indicated above

done

STRML:
Thanks!

n1ghtmare:
> @n1ghtmare
> 
> > Maybe we can work on the documentation and refer users to it for this specific edge case.
> 
> I've already done it (https://github.com/STRML/react-grid-layout/pull/1010/files#diff-04c6e90faac2675aa89e2176d2eec7d8R317)
> 
> > Meanwhile we can keep an eye at the firefox bug tracker for a possible fix.
> 
> This issue (https://bugzilla.mozilla.org/show_bug.cgi?id=568313) was created 10 years ago so I think they won't fix it in the nearest future :)

Sorry, I didn't see that. Great job man, glad it got merged.

Issue #1011, opened 2019-09-26
getDerivedStateFromProps not updating the layout properly


Lagily:
getDerivedStateFromProps seems to not update the layout property, when changing the layout property with the newest React version seemingly nothing happens. I have debugged it and it seems like nextProps.layout and prevState.propLayout are always identical in the getDerivedStateFromProps hook. This might be due to some recent bug fixes by the react team: https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html

Pull request will follow, please have a look.

zaykaalexander:
We checked this bug on all versions from 16.4 to the current one. We did not meet anything that could disrupt the grid, if you change prop "layout". If you have more information, then share it with us so that we can more accurately reproduce the error.

Lagily:
hey, thanks for checking it out, please check this codesandbox

https://codesandbox.io/s/wizardly-hugle-s7x0h?fontsize=14

After 3 seconds the static flag is set to false and a new state is set, therefore the GridLayout receives a new layout props. However, due to the faulty check for layout change, the new layout is not loaded and the 'a' in the example remains static. The same code run with the changes I made in my pull request results in the layout updating after 3 seconds and therefore the 'a' becomes non static correctly.

Let me know if it's somewhat unclear.

Expected behaviour: After 3 seconds the item 'a' in the grid layout becomes non static.
Actual behaviour: nothing happens, even though the state is updated and the new layout is passed as a prop to the GridLayout

al-fyodorov:
@Lagily, hey, your example not exact, because array copied only, and not copied items from array which are using by link, so `nextProps.layout` and `prevProps.propsLayout` is equal in `getDerivedStateFromProps`.

Fixed example: https://codesandbox.io/s/quirky-roentgen-jc7i1

Me it seems, this not library problem :slightly_smiling_face:

PR #1012, opened 2019-09-26
propsLayout changed to layout for layout update check

Lagily:
fix for https://github.com/STRML/react-grid-layout/issues/1011

### DIFF
--- a/lib/ReactGridLayout.jsx
+++ b/lib/ReactGridLayout.jsx
@@ -315,7 +315,7 @@ export default class ReactGridLayout extends React.Component<Props, State> {
     // Legacy support for compactType
     // Allow parent to set layout directly.
     if (
-      !isEqual(nextProps.layout, prevState.propsLayout) ||
+      !isEqual(nextProps.layout, prevState.layout) ||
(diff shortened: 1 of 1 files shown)


lib/ReactGridLayout.jsx
```diff
@@ -315,7 +315,7 @@ export default class ReactGridLayout extends React.Component<Props, State> {
     // Legacy support for compactType
     // Allow parent to set layout directly.
     if (
-      !isEqual(nextProps.layout, prevState.propsLayout) ||
+      !isEqual(nextProps.layout, prevState.layout) ||
```
daynin: we can't compare layout from props and from state here because of `synchronizeLayoutWithChildren`, which will create `synchronized` layout (with different fields) so `isEqual(nextProps.layout, prevState.layout)` will be always false. In this case we want to compare new layout from props and the previous one. And it work now pretty good

I tried to figure out what is the problem here https://github.com/STRML/react-grid-layout/issues/1011 but I couldn't

Could you describe the issue with an example please?

Lagily:
were you able to check?

PR #1016, opened 2019-09-30
added an ability to use different margins for different breakpoints. #964

### LINKED ISSUE #964: Question: Is it possible to have different margin and containerPadding config for different Breakpoints?
Hi,
I have a requirement where I need to have slightly different grid config(margin and containerPadding) in mobile compared to other viewports. Is this something I can achieve in RGL?

serebrennikovalex:
Added an ability to use different margins, containerPaddings for different breakpoints.

Fixes #964

### DIFF
--- a/README.md
+++ b/README.md
@@ -359,12 +359,13 @@ breakpoints: ?Object = {lg: 1200, md: 996, sm: 768, xs: 480, xxs: 0},
 // # of cols. This is a breakpoint -> cols map, e.g. {lg: 12, md: 10, ...}
 cols: ?Object = {lg: 12, md: 10, sm: 6, xs: 4, xxs: 2},
 
-// # of margin. This is a breakpoint -> margin map, e.g. {lg: [10, 10], md: [10, 10], ...} or [10 , 10]
-margin: [number, number] | ?Object = [10, 10],
+
+// margin (in pixels). Can be specified either as horizontal and vertical margin, e.g. `[10, 10]` or as a breakpoint -> margin map, e.g. `{lg: [10, 10], md: [10, 10], ...}.
+margin: [number, number] | {[breakpoint: $Keys<breakpoints>]: [number, number]}
 
 
-// # of containerPadding. This is a breakpoint -> containerPadding map, e.g. {lg: [10, 10], md: [10, 10], ...} or [10 , 10]
-containerPadding: [number, number] | ?Object = [10, 10],
+// containerPadding (in pixels). Can be specified either as horizontal and vertical padding, e.g. `[10, 10]` or as a breakpoint -> containerPadding map, e.g. `{lg: [10, 10], md: [10, 10], ...}.
+containerPadding: [number, number] | {[breakpoint: $Keys<breakpoints>]: [number, number]}
 
 
 // layouts is an object mapping breakpoints to layouts.
--- a/README.md
+++ b/README.md
@@ -359,12 +359,12 @@ breakpoints: ?Object = {lg: 1200, md: 996, sm: 768, xs: 480, xxs: 0},
 // # of cols. This is a breakpoint -> cols map, e.g. {lg: 12, md: 10, ...}
 cols: ?Object = {lg: 12, md: 10, sm: 6, xs: 4, xxs: 2},
 
-// # of margin. This is a breakpoint -> margin map, e.g. {lg: [10, 10], md: [10, 10], ...}
-margin: [number, number] | ?Object = [10, 10] | {lg: [10, 10], md: [10, 10], sm: [10, 10], xs: [10, 10], xxs: [10, 10]},
+// # of margin. This is a breakpoint -> margin map, e.g. {lg: [10, 10], md: [10, 10], ...} or [10 , 10]
+margin: [number, number] | ?Object = [10, 10],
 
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -359,6 +359,14 @@ breakpoints: ?Object = {lg: 1200, md: 996, sm: 768, xs: 480, xxs: 0},
 // # of cols. This is a breakpoint -> cols map, e.g. {lg: 12, md: 10, ...}
 cols: ?Object = {lg: 12, md: 10, sm: 6, xs: 4, xxs: 2},
 
+// # of margin. This is a breakpoint -> margin map, e.g. {lg: [10, 10], md: [10, 10], ...}
+margin: ?Object = {lg: [10, 10], md: [10, 10], sm: [10, 10], xs: [10, 10], xxs: [10, 10]},
```
daynin: we definitely should keep backward compatibility so we should change that type (in my opinion) like this:

```
[number, number] | ?Object
```

and we should do it for containerPadding as well

README.md
```diff
@@ -359,6 +359,14 @@ breakpoints: ?Object = {lg: 1200, md: 996, sm: 768, xs: 480, xxs: 0},
 // # of cols. This is a breakpoint -> cols map, e.g. {lg: 12, md: 10, ...}
 cols: ?Object = {lg: 12, md: 10, sm: 6, xs: 4, xxs: 2},
 
+// # of margin. This is a breakpoint -> margin map, e.g. {lg: [10, 10], md: [10, 10], ...}
+margin: [number, number] | ?Object = [10, 10] | {lg: [10, 10], md: [10, 10], sm: [10, 10], xs: [10, 10], xxs: [10, 10]},
```
daynin: here we should describe default value, it can't be an array and an object in the same time. In the code it's an object for now:

```typescript
margin: {
      lg: [10, 10],
      md: [10, 10],
      sm: [10, 10],
      xs: [10, 10],
      xxs: [10, 10]
    },
```

But I think it'll be better to use an array here (e. g. [10, 10]), because it gives the same result.

README.md
```diff
@@ -359,6 +359,14 @@ breakpoints: ?Object = {lg: 1200, md: 996, sm: 768, xs: 480, xxs: 0},
 // # of cols. This is a breakpoint -> cols map, e.g. {lg: 12, md: 10, ...}
 cols: ?Object = {lg: 12, md: 10, sm: 6, xs: 4, xxs: 2},
 
+// # of margin. This is a breakpoint -> margin map, e.g. {lg: [10, 10], md: [10, 10], ...}
```
daynin: add pls information about backward compatibility: `e.g. {lg: [10, 10], md: [10, 10], ...}` -> `e.g. {lg: [10, 10], md: [10, 10], ...} or [10 , 10]`

lib/ResponsiveReactGridLayout.jsx
```diff
@@ -238,22 +254,41 @@ export default class ResponsiveReactGridLayout extends React.Component<
...
   }
 
+  /**
+   * Get a value of margin or containerPadding.
+   *
+   * @param  {Array | Object} param Margin | containerPadding, e.g. [10, 10] | {lg: [10, 10], ...}.
+   * @param  {String} breakpoint   Breakpoint: lg, md, sm, xs and etc.
+   * @return {Array}
+   */
+  getMarginPaddingValue = (param: any, breakpoint: string) => {
```
daynin: I think it'll be better to extract this method into a separate function in this file. It doesn't seem like a member of the class and maybe we should rename it: 'getMarginPaddingValue' -> 'getIndentationValue'

README.md
```diff
@@ -359,6 +359,14 @@ breakpoints: ?Object = {lg: 1200, md: 996, sm: 768, xs: 480, xxs: 0},
 // # of cols. This is a breakpoint -> cols map, e.g. {lg: 12, md: 10, ...}
 cols: ?Object = {lg: 12, md: 10, sm: 6, xs: 4, xxs: 2},
 
+// # of margin. This is a breakpoint -> margin map, e.g. {lg: [10, 10], md: [10, 10], ...} or [10 , 10]
```
STRML: This comment is a little unclear. 

Suggest:

``` js
// margin (in pixels). Can be specified either as horizontal and vertical margin, e.g. `[10, 10]` or as a breakpoint -> margin map, e.g. `{lg: [10, 10], md: [10, 10], ...}.
margin: [number, number] | {[breakpoint: $Keys<breakpoints>]: [number, number]}
```

lib/ResponsiveReactGridLayout.jsx
```diff
@@ -20,6 +20,18 @@ import type { Layout } from "./utils";
...
 
+/**
+ * Get a value of margin or containerPadding.
+ *
+ * @param  {Array | Object} param Margin | containerPadding, e.g. [10, 10] | {lg: [10, 10], ...}.
+ * @param  {String} breakpoint   Breakpoint: lg, md, sm, xs and etc.
+ * @return {Array}
+ */
+
+function getIndentationValue(param: any, breakpoint: string) {
```
STRML: Unexpected `any` type here

daynin:
@STRML i think we should merge it

STRML:
LGTM, thanks.

PR #1018, opened 2019-09-30
Remove the warning when the compactType is set to null

albandewilde:


albandewilde:
I'm closing this PR because it was opened long time ago and it may be obsolete now.

PR #1019, opened 2019-10-01
fixed RangeError: Maximum call stack size exceeded.

### LINKED ISSUE #941: Dragging an element over a static element causes RangeError: Maximum call stack size exceeded
I've noticed has been mentioned before in threads but there doesn't seem to be a fix.

When you drag an element over a static element on a board which has 'preventCollision: false' and 'compactType: null' you get "RangeError: Maximum call stack size exceeded".

Error produced:
Uncaught RangeError: Maximum call stack size exceeded
    at bundle.js:6555
    at Array.sort (<anonymous>)
    at sortLayoutItemsByRowCol (bundle.js:6555)
    at sortLayoutItems (bundle.js:6551)
    at moveElement (bundle.js:6427)
    at moveElementAwayFromCollision (bundle.js:6496)
    at moveElement (bundle.js:6451)
    at moveElementAwayFromCollision (bundle.js:6496)
    at moveElement (bundle.js:6451)
    at moveElementAwayFromCollision (bundle.js:6496)

From then it just continually produces moveElement and moveElementAwayFromCollision errors

Here's a sandbox of it - https://codesandbox.io/embed/reactgridlayout-isfjp

serebrennikovalex:
Fixed RangeError: Maximum call stack size exceeded when dragging an element over a static element causes.

The point of the solution is that when we drag an element over a static one, a collision occurs and we have to interrupt the operation.

Fixes #941

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -406,9 +406,7 @@ export function moveElement(
   for (let i = 0, len = collisions.length; i < len; i++) {
     const collision = collisions[i];
     log(
-      `Resolving collision between ${l.i} at [${l.x},${l.y}] and ${
-        collision.i
-      } at [${collision.x},${collision.y}]`
+      `Resolving collision between ${l.i} at [${l.x},${l.y}] and ${collision.i} at [${collision.x},${collision.y}]`
     );
 
     // Short circuit so we can't infinite loop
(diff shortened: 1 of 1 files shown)


lib/utils.js
```diff
@@ -458,7 +456,12 @@ export function moveElementAwayFromCollision(
   const compactH = compactType === "horizontal";
   // Compact vertically if not set to horizontal
   const compactV = compactType !== "horizontal";
-  const preventCollision = false; // we're already colliding
+  let preventCollision = false; // we're already colliding
```
daynin: it can be done a little bit simpler:

```typescript
const preventCollision = collidesWith.static
```

and, I think, we should change this comment `we're already colliding` -> `we're already colliding (not for static items)` and remove comment `when preventCollision = false and element is static then we have RangeError: Maximum call stack size exceeded`, it's better to place this comment into the commit message

serebrennikovalex: I agree, thank you.

STRML:
Thanks for the contribution!

Issue #1028, opened 2019-10-04
Input focus lost when updating "isResizable" or "isDraggable"


sopretty:
- [x] Bug
- [ ] Feature request
- [ ] Question

Note: I can only reproduce this issue on Chrome.

In my example, I want to set the `isResizable` or `isDraggable` props to false once I start typing on a text input. But when the values change, I can't keep typing anymore in the input, I have to focus the input again.

Here's a sandbox example (remember to try on Chrome)
https://codesandbox.io/s/pedantic-faraday-cedgd

**Steps to reproduce:**
1. The grid items are resizable and draggable
2. Focus input
3. Enter one letter
4. The grid items are not resizable and draggable | The focus of the input is lost

ElaiNili:
I also had the same problem, I resolved it in this way. 

```
static get defaultProps() {
    return {
      draggableCancel: '.nondrag',
    };
  }
```
_______________
```
<ResponsiveReactGridLayout
          { ...this.props} >
```
______________
`<div className="nondrag">`

NOTICE: do not forget to pass props into ResponsiveReactGridLayout!

palashkaria:
I'm facing the same issue when using an input to filter draggable list. The fix provided in #1113 worked in my case; hoping to see a release soon!

PR #1030, opened 2019-10-07
#1028 fix for input focus lost

al-fyodorov:
Passing `react-draggable` props for disable user select hack in `GridItem`, because is him broken focus on input field.

https://github.com/STRML/react-grid-layout/issues/1028


palashkaria:
This seems stale, we can close this in favour of #1113

Issue #760, opened 2018-03-18
Drag and Drop performance in IE11


ConfigurableUIComponents:
We're experiencing very slow dragging and dropping in IE11; including in the demos linked in the official documentation. any tips on how we can improve this? analyze it? will be happy to contribute.

PR #1032, opened 2019-10-08
Fix freezing on drag ie11

zaykaalexander:
https://github.com/STRML/react-grid-layout/issues/760

1) Add shouldComponentUpdate for ReactGridLayout
2) Add shouldComponentUpdate for GridItem

example or test case.

### DIFF
--- a/lib/ReactGridLayout.jsx
+++ b/lib/ReactGridLayout.jsx
@@ -14,7 +14,8 @@ import {
   synchronizeLayoutWithChildren,
   validateLayout,
   getAllCollisions,
-  noop
+  noop,
+  pick
 } from "./utils";
 import GridItem from "./GridItem";
 import type {
@@ -95,6 +96,33 @@ const compactType = (props: Props): CompactType => {
 
   return verticalCompact === false ? null : compactType;
 };
+
+/**
+ * Array of keys for sCU
+ */
+const pickPropsKeys = [
+  "className",
+  "style",
+  "width",
+  "autoSize",
+  "cols",
+  "draggableCancel",
+  "draggableHandle",
+  "verticalCompact",
+  "compactType",
+  "layout",
+  "margin",
+  "containerPadding",
+  "rowHeight",
+  "maxRows",
+  "isDraggable",
+  "isResizable",
+  "isDroppable",
+  "preventCollision",
+  "useCSSTransforms",
+  "transformScale",
+  "droppingItem"
+];
 
 /**
  * A reactive, fluid grid layout with draggable, resizable components.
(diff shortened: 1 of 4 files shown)


lib/ReactGridLayout.jsx
```diff
@@ -348,6 +348,13 @@ export default class ReactGridLayout extends React.Component<Props, State> {
     return null;
   }
 
+  shouldComponentUpdate(nextProps: Props, nextState: State) {
+    return (
+      !isEqual(this.props, nextProps) ||
```
daynin: we really want to compare entire props object here? Maybe we can compare certain props only?

zaykaalexander: With any changes in props, we must rerender the component. We only check the input and only some internal state

lib/ReactGridLayout.jsx
```diff
@@ -348,6 +376,15 @@ export default class ReactGridLayout extends React.Component<Props, State> {
     return null;
   }
 
+  shouldComponentUpdate(nextProps: Props, nextState: State) {
+    return (
+      !isEqual(
```
daynin: I think we can revert this logic and use lodash.omit for removing properties which we won't compare. Like this:

```typescript
!isEqual(omit(this.props, propsToExclude), omit(nextProps, propsToExclude))
```

I suggested to revert this logic because it seems like excluded props list will be shorter than included one. But if I'm wrong, you can use your function `iclude`, but i think we should rename it into `pick`

zaykaalexander: Function renamed to "pick", As for me, it’s better to include only parameters you need

STRML:
Any profiles showing actual perf improvements here? `calcPosition` is fast but not *that* fast, it could end up being more expensive overall to have such an involved SCU.

zaykaalexander:
Perfomance test
Google Chrome Perfomance Screenshots:
Before: http://s.csssr.ru/UL3G4FBBJ/Before.png
After: http://s.csssr.ru/UL3G4FBBJ/After.png

zaykaalexander:
Video perfomance
Before: http://s.csssr.ru/UL3G4FBBJ/Before.mp4
After: http://s.csssr.ru/UL3G4FBBJ/After.mp4

daynin:
LGTM

daynin:
@STRML what do you think?

STRML:
Something is very odd about the way items are animating in the `After` example, as if they are animating from an origin on the top-left. May be a regression there.

daynin:
@STRML let's release the new version without this PR. And include this PR in the next release (0.17.1 for example)

STRML:
0.17.0 now released - on to 0.17.1.

n1ghtmare:
I'm really excited about the release. Good job everyone!

zaykaalexander:
> Something is very odd about the way items are animating in the `After` example, as if they are animating from an origin on the top-left. May be a regression there.

The problem fixed

STRML:
Taking a look at this now for inclusion in 0.18.0.

STRML:
See #1123

STRML:
Merged in #1123

PR #1033, opened 2019-10-09
fix: remove dropping placeholder on drag leave

daynin:
remove the placeholder when a droppable node leaved the layout (an example of the bug: https://s.csssr.ru/U286BQJEP/20191009113052.mp4)


daynin:
@STRML what about merging this PR (or maybe I should fix something here)?

STRML:
Looks good, thanks a lot.

PR #1035, opened 2019-10-09
chore(actions): add action for removing stale issues and RPs

daynin:
It'll help us to control stale issues and RPs and in current situation with issues and PRs it'll help us to filter really important ones from irrelevant


STRML:
Yes that's quite good, thank you.

Issue #1039, opened 2019-10-10
Unify prettier config


STRML:
@daynin https://github.com/STRML/react-grid-layout/pull/1013#pullrequestreview-299891540

PR #1041, opened 2019-10-11
Unify prettier config

mei33:
#1039 
This PR explicitly specifies to use default prettier config (which is already in use).


daynin:
LGTM

Issue #901, opened 2019-01-23
MouseClick on GridItem let move the Item - Why? :|


PirolA:
Here is the link to reproduce the bug: https://codesandbox.io/s/r0kqjk8xq4

You have to open the resulting layout in a new window ("Open in new window") then resize the newly opened window to the half. After that you have to click on a griditem which has no neighbor to the right -> Here you go, the item jumps to the right.

PirolA:
it has nothing to do with the fact that x is greater than y. The bug always occurs when the size of the windows has reached a multiple of x, then the problem is there. At an x-value of 70px, the bug appears starting from a window size of about 21 times, that is about window width of 1500px.

bartoszkrawczyk2:
I'm experiencing the same issue. Any update on this? Have you managed to solve it @PirolA ?

bartoszkrawczyk2:
@PirolA for me setting the same value ([20, 20]) for `containerPadding` as `margin` on `<GridLayout>` solved this issue.

daynin:
maybe it relates to https://github.com/STRML/react-grid-layout/issues/866

PR #1044, opened 2019-10-15
#901 Forbid change layout on click without drag

mei33:
It resolves https://github.com/STRML/react-grid-layout/issues/901 and does not allow to change layout until layout item was dragged. BTW, it is still doesn't clear why that bug wasn't reproducible on examples included in library.


STRML:
Thanks!

Issue #478, opened 2017-02-10
[Bug] - grid item isDraggable does not override global setting


johnsiu2016:
http://www.webpackbin.com/4k92w68dz

```javascript
render() {
    return (
      <div>
        <div>Current Breakpoint: {this.state.currentBreakpoint} ({this.props.cols[this.state.currentBreakpoint]} columns)
        </div>
        <button onClick={this.onNewLayout}>Generate New Layout</button>
        <ResponsiveReactGridLayout
          {...this.props}
          layouts={this.state.layouts}
          onBreakpointChange={this.onBreakpointChange}
          onLayoutChange={this.onLayoutChange}
          // WidthProvider option
          measureBeforeMount={true}
          // I like to have it animate on mount. If you don't, delete `useCSSTransforms` (it's default `true`)
          // and set `measureBeforeMount={true}`.
          useCSSTransforms={this.state.mounted}
          isDraggable={false}
          isResizable={false}>
          {this.generateDOM()}
        </ResponsiveReactGridLayout>
      </div>
    );
  }
```
```javascript
function generateLayout() {
  return _.map(_.range(0, 25), function (item, i) {
    var y = Math.ceil(Math.random() * 4) + 1;
    return {
      x: _.random(0, 5) * 2 % 12,
      y: Math.floor(i / 6) * y,
      w: 2,
      h: y,
      i: i.toString(),
      static: true,
      isDraggable: true,
      isResizable: true
    };
  });
}
```

When I set the layout to isDraggable false
```javascript
          isDraggable={false}
          isResizable={false}>
```

and the grid item to isDraggable true
```javascript
      static: true,
      isDraggable: true,
      isResizable: true
```

The result is isDraggable false
![image](https://cloud.githubusercontent.com/assets/17636906/22832376/be4ac98a-efe8-11e6-968b-dd89357b8271.png)

However, the document wrote:
Any <GridItem> properties defined directly will take precedence over globally-set options. For example, if the layout has the property isDraggable: false, but the grid item has the prop isDraggable: true, the item will be draggable.

Is it right?

In reactGridLayout.jsx,

```javacript
    // Parse 'static'. Any properties defined directly on the grid item will take precedence.
    const draggable = Boolean(!l.static && isDraggable && (l.isDraggable || l.isDraggable == null));
    const resizable = Boolean(!l.static && isResizable && (l.isResizable || l.isResizable == null));
```

When isDraggable is false, the whole statement is false.

PR #1050, opened 2019-10-23
Fix overriding global options for dragging and resizing grid item

polikashina:
It resolves [#478 ](https://github.com/STRML/react-grid-layout/issues/478) and allows local options to override global ones for resizing and dragging grid item

### DIFF
--- a/lib/ReactGridLayout.jsx
+++ b/lib/ReactGridLayout.jsx
@@ -2,6 +2,7 @@
 import React from "react";
 import PropTypes from "prop-types";
 import isEqual from "lodash.isequal";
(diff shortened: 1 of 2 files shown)


lib/ReactGridLayout.jsx
```diff
@@ -2,6 +2,7 @@
 import React from "react";
 import PropTypes from "prop-types";
 import isEqual from "lodash.isequal";
+import isNil from "lodash.isnil";
```
daynin: i think we can do it without installing one more dependency

STRML:
Superseded in #1060.

Issue #1046, opened 2019-10-15
TransformScale is not applied properly while dragging an element


MrDockal:
![captured](https://user-images.githubusercontent.com/8297716/66874768-bcd52280-f008-11e9-883a-7bca6670e81c.gif)

The dragging element is slower (or faster) than the mouse cursor when parent element of `react-grid-layout` is scaled.
```
<div style={{transform: 'scale(0.5)'}}>
    <ResponsiveReactGridLayout
        transformScale={0.5}
        {...this.props}
    />
</div>
```
I am aware this is an issue with [react-draggable](https://github.com/mzabriskie/react-draggable)

PR #1054, opened 2019-10-24
TransformScale is not applied properly while dragging an element #1046

serebrennikovalex:
it's a supplement to the PR https://github.com/mzabriskie/react-draggable/pull/439

Fixes: #1046


serebrennikovalex:
@STRML @daynin please review these changes.

Issue #1056, opened 2019-10-28
0.17.0 is checking for navigator even if there is no window


alexvision:
The Firefox check here should only occur in a context where window exists: https://github.com/STRML/react-grid-layout/blob/master/lib/ReactGridLayout.jsx#L100

DylanJu:
Here is the same error for me. I fixed temporarily by fixing the version down to 0.16.6.

alexvision:
This change fixes it for me if you wanted to try it  https://github.com/STRML/react-grid-layout/issues/1057

PR #1057, opened 2019-10-28
Protect from non-existent or bad implementations of `navigator`

alexvision:
This was an issue for my usage as the package is imported in a node context, and this navigator check is done at the top level of this package, rather than after rendering the grid. So window is undefined at the point that it is included initially. 

This fixes the issue I created here: https://github.com/STRML/react-grid-layout/issues/1056


STRML:
Note: Labeler failure due to https://github.com/TimonVS/pr-labeler-action/issues/25

STRML:
LGTM, thanks.

PR #1060, opened 2019-10-29
fix(GridItem): ensure locally set isDraggable/isResizable overrides globals

### LINKED ISSUE #478: [Bug] - grid item isDraggable does not override global setting

STRML:
This also makes static items resizable/draggable if locally set, which is
a nice option but will break backcompat.

Supersedes #1050
Fixes #478

### DIFF
--- a/README.md
+++ b/README.md
@@ -287,7 +287,7 @@ containerPadding: ?[number, number] = margin,
 // if you like.
 rowHeight: ?number = 150,
 
-// Configuration of a dropping element. Dropping element is a "virtual" element 
+// Configuration of a dropping element. Dropping element is a "virtual" element
 // which appears when you drag over some element from outside.
 // It can be changed by passing specific parameters:
 //  i - id of an element
@@ -311,9 +311,9 @@ transformScale: ?number = 1,
 // dragged over.
 preventCollision: ?boolean = false;
 
-// If true, droppable elements (with `draggable={true}` attribute) 
+// If true, droppable elements (with `draggable={true}` attribute)
 // can be dropped on the grid. It triggers "onDrop" callback
-// with position and event object as parameters. 
+// with position and event object as parameters.
 // It can be useful for dropping an element in a specific position
 //
 // NOTE: In case of using Firefox you should add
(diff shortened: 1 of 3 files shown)


STRML:
This backcompat break would necessitate a major bump.

PR #1070, opened 2019-11-08
add-id-prop-to-react-grid-layout

carole-bennett:
In order to use more than one react-grid-layout on a page, such as in a scenario where layouts could be nested, an ID prop is needed for dynamic generation, and drag-drop layout functionality. 

I have a fully functional layout editor, using nested react-grid-layouts, enabled by the addition of this additional prop.  I'd really prefer not to have to install my own fork to to support it, and it's a fairly straightforward addition.

### DIFF
--- a/lib/ReactGridLayout.jsx
+++ b/lib/ReactGridLayout.jsx
@@ -50,6 +50,7 @@ type State = {
 };
 
 export type Props = {
+  id: string,
   className: string,
   style: Object,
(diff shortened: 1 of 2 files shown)


STRML:
This doesn't actually use the `id` prop, so I'm curious as to what the utility of this is or why you wouldn't simply wrap each layout in a `<div>` with an `id`.

Issue #1083, opened 2019-11-20
Cols update doesn't reflect changes in version 0.17.1


minotogna:
Hello,

I noticed that in version 0.17.1 if cols is dynamically update, the grid doesn't reflect the change.
It used to work in version 0.16.6.

I have create a test case here:
https://codepen.io/drusus-uttarayananda-minor/pen/KKKEKvY

Thank you.
Mino

lebbe:
I also experience this problem, updating cols doesn't update the GridLayout.

lebbe:
I found kind of like a workaround. We don't need the moved flag in the layout, so I just flip this boolean every time cols is updated, and the grid is redrawn with the new cols.

```
this.setState({
	layout: [{...a, moved: !a.moved}, ...rest],
	cols: cols
});
```

PR #1090, opened 2019-11-27
Fix handling of width changes

gregbty:
The width change event was ignoring dynamic changes to breakpoints and columns because it was checking current props against itself. This bug was introduced when the library refactored the use of the deprecated componentWillRecieveProps in favor of componentDidUpdate. This should resolve the bug reported in #1083.


steinbergh:
would adore this being merged in, the responsive layout is essentially unusable without this fix

STRML:
Thanks, this is a pretty serious bug - I apologize for not merging this in sooner.

Issue #1068, opened 2019-11-06
onDrop parameters / Drag from outside example


raymond-ong:
Hello,
I am just exploring this library's Drag from Outside example, and also a React newbie so pardon my question.

So the example simply spits out an alert message when the outside item is dragged into the layout.
So I simply want to "formally" add the outside item into the layout.

If my understanding is correct, i need to call setState() for the layout containing the "new" layout.
By "new" layout, i mean:

a. it contains the new outside item
b. updated XY coordinates of existing items that may have been pushed over by the new outside item.

-> If I simply call setState() by providing "a" above, without performing "b", the new outside item is being added into "random" coordinates.

So my question is whether this "new" layout is readily available as a props/callback?
I have noted that onDragStop() is not getting fired for outside items.
But onDrag() is fired, and its "layout" parameter actually contains the "new" layout I mentioned above.

**One option** I tried, which works but NOT Efficient, is to add a temporary state variable that stores the "new" layout during onDrag, and use it later when onDrop is called.

**The 2nd option** I tried is to modify the library's ReactGridLayout.js file.
so inside 

> _defineProperty(_assertThisInitialized(_this), "onDrop", function ()

I added the "layout" variable as the callback parameter, like so:

`      _this.props.onDrop({
        x: x,
        y: y,
        w: w,
        h: h
      },
      layout // added by me!
      );`

If my 2nd option is OK, then I hope this gets checked in to the library repo.
Or is there any other way?

Actually I hope the example can be updated to not just spit out an alert message, but also to add the new outside item into the layout :-)

Jacob-Allen:
@raymond-ong 
I ran into a need for this as well. Almost every time an item was dropped into the grid it would have caused a collision so many of the other elements move around as well, so simply getting the coordinates of the new item was not sufficient enough to accurately know the new layout.

What I did to get around this was ref up ReactGridLayout and I just manually inspected the state.layout of the child class so I did not have to add any new code to the package. I would not call this best practice, but it does get around this limitation for now.

See my example below, I've provided both a class and a functional component as reference.

https://gist.github.com/Jacob-Allen/b80f9ff07244054b059bc6845af72afa

raymond-ong:
Hi @Jacob-Allen,
Thank you for your suggestion and gist!
I also don't want to change the library code if possible.

Your suggestion works for a normal GridLayout. The ref'd child's state contains the layout field.
[...]

PR #1101, opened 2019-12-11
Adding layout param to onDrop callback

cheryllaird:
This change passes the new layout into the onDrop callback, this will allow us to handle the fact that adding a new item may have caused other items in the grid to shift.

https://github.com/STRML/react-grid-layout/issues/1068

### DIFF
--- a/README.md
+++ b/README.md
@@ -353,7 +353,7 @@ onResize: ItemCallback,
 // Calls when resize is complete.
 onResizeStop: ItemCallback,
 // Calls when some element has been dropped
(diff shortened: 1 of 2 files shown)


raymond-ong:
Thanks Cherry. Hopefully this will be included in the next release.

cheryllaird:
Commenting so the PR does not get closed...

STRML:
I thought #1169 was more consistent with other callbacks, so I merged that arity instead. Thanks for the contribution!

PR #1113, opened 2019-12-30
Add `enableUserSelectHack` to ReactGridLayout's props

fxxjdedd:
This property will cause some performance issues. I will manually disable user-select to optimize performance, so I need a prop here.

### DIFF
--- a/lib/GridItem.jsx
+++ b/lib/GridItem.jsx
@@ -51,6 +51,7 @@ type Props = {
   // Draggability
   cancel: string,
   handle: string,
+  enableUserSelectHack: boolean,
 
   x: number,
   y: number,
@@ -148,6 +149,7 @@ export default class GridItem extends React.Component<Props, State> {
     handle: PropTypes.string,
     // Selector for draggable cancel (see react-draggable)
     cancel: PropTypes.string,
+    enableUserSelectHack: PropTypes.bool,
     // Current position of a dropping element
     droppingPosition: PropTypes.shape({
       e: PropTypes.object.isRequired,
@@ -164,7 +166,8 @@ export default class GridItem extends React.Component<Props, State> {
     minW: 1,
     maxH: Infinity,
     maxW: Infinity,
-    transformScale: 1
+    transformScale: 1,
+    enableUserSelectHack: true
   };
 
   state: State = {
@@ -381,6 +384,7 @@ export default class GridItem extends React.Component<Props, State> {
           (this.props.cancel ? "," + this.props.cancel : "")
         }
         scale={this.props.transformScale}
+        enableUserSelectHack={this.props.enableUserSelectHack}
       >
         {child}
       </DraggableCore>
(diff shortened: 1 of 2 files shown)


fxxjdedd:
Adding just one attribute looks stupid, or is there a better way to pass in react-draggable props?

nikhil-varma:
@fxxjdedd Is there any update on this? I am trying to use this change and this issue still seems to remain..  I am passing down `enableUserSelectHack={false}` from the file I am using `<GridLayout />`...

fxxjdedd:
> @STRML @fxxjdedd Is there any update on this? I am trying to use these changes made but this issue still seems to remain.. I am passing down `enableUserSelectHack={false}` from the file I am using `<GridLayout />`...

Not yet. Can this commit be merged? @STRML

nikhil-varma:
@fxxjdedd Can we also confirm the changes are not quite solving the issue... Or am I doing something wrong?

fxxjdedd:
This PR was originally used to solve this problem: before dragging, react-draggable will dynamically change the style of HTML tags, thus disabling user select. In most cases, it works well, but reflow can cause poor page performance when there are many GridItems. Therefore, by passing on this property, and handed over to maintain when to disable user. 

This PR works well for me, but for input focus loss, I'm not sure it works.

nikhil-varma:
@STRML Can we get this merged soon?

nikhil-varma:
@fxxjdedd I realized that this PR adds support to take `enableUserSelectHack` props in the `placeholder()` function inside GridItem. Shouldn't `processGridItem` also have the support for this? That fix locally seem to have resolved the focus issue for good.

fxxjdedd:
@nikhil-varma Thanks, i have fixed it

STRML:
Would prefer we add `draggableOpts` rather than whitelist a single `<Draggable>` prop here.

fxxjdedd:
@STRML 

I made some changes according to your suggestion.

But it is difficult to make it perfect, since we have extracted the properties (cancel, handle, ...) of `draggableOpts` as props of GridItem, it is not perfect to add another `draggableOpts` to GridItem.

Maybe this pr should be closed until someone finds a better way.

STRML:
I think for now passing `draggableOpts` is fine, then we can consider removing the old `handle/cancel` options in a later major.

lib/GridItem.jsx
```diff
@@ -348,6 +351,7 @@ export default class GridItem extends React.Component<Props, State> {
           (this.props.cancel ? "," + this.props.cancel : "")
         }
         scale={this.props.transformScale}
+        {...this.props.draggableOpts}
```
STRML, line 354: Let's put this at the beginning of the props list so that the other props will override them, in case someon inadvertently sets `handle/cancel` et al.

lib/GridItem.jsx
```diff
@@ -393,7 +398,8 @@ export default class GridItem extends React.Component<Props, State> {
     return (
       <Resizable
         draggableOpts={{
-          disabled: !isResizable
+          disabled: !isResizable,
+          ...draggableOpts
```
STRML, line 402: And same here. User shouldn't be able to set `disabled` this way.

STRML (commented):
Would also be great to insert a test if you have the time.

fxxjdedd:
Should we also add `draggableOpts` to `LayoutItem` to control a single GridItem?

Fiv5:
Is there any update on this?

fxxjdedd:
it's ok with self-test.

Fiv5:
@STRML could you deal with this PR please?

petdomaa100:
Any updates?

bliitzkrieg:
Any update on this?

STRML:

PR #1115, opened 2020-01-07
Feature: add initialItemLayout option

eric-burel:
Hi,

This PR add a new option to configure the initial item layout, used as the default layout for new grid items. In some case a 1x1 block is rather small and not suited as the default size of a block. User can configure `initialLayoutItem: { w: 4, h: 4}` for example.

However I have some trouble testing this in my apps. Using `npm link` create a double React version issue (I use 15.x in my app), so I have to somehow copy paste the code. But I've managed to run a small local test and it was ok.

### DIFF
--- a/README.md
+++ b/README.md
@@ -279,6 +279,11 @@ compactType: ?('vertical' | 'horizontal') = 'vertical';
 // {i: string, x: number, y: number, w: number, h: number}
 layout: ?array = null, // If not provided, use data-grid props on children
 
+// Initial layout for new items that do not yet possess a layout
+// changing height and width is especially useful if items are too small:
+// { w: 4, h: 4} will create items of 4 blocks per 4 blocks
+initialLayoutItem?: { w: number, h: number, x: number, y: number}
+
 // Margin between items [x, y] in px.
 margin: ?[number, number] = [10, 10],
 
--- a/lib/ReactGridLayout.jsx
+++ b/lib/ReactGridLayout.jsx
@@ -60,6 +60,7 @@ export type Props = {
   verticalCompact: boolean,
   compactType: CompactType,
   layout: Layout,
+  initialLayoutItem: $Shape<LayoutItem>,
   margin: [number, number],
   containerPadding: [number, number] | null,
   rowHeight: number,
@@ -160,6 +161,13 @@ export default class ReactGridLayout extends React.Component<Props, State> {
       if (layout === undefined) return;
       validateLayout(layout, "layout");
     },
+
+    initialLayoutItem: PropTypes.shape({
+      x: PropTypes.number,
+      y: PropTypes.number,
+      h: PropTypes.number,
+      w: PropTypes.number
+    }),
 
     //
     // Grid Dimensions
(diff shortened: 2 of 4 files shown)


STRML:
I don't think this makes sense to add to the library. If you are adding a new item, you can simply use the `data-grid` property or mutate your layout to set a default size.

I don't want to unnecessarily increase the size of this library to fit every use case.

eric-burel:
Thanks for your answer. Do you think it would be possible to add only a partial data-grid? 

Here basically I only want to set the with/height for children that are not yet in the layout, but I can't tell yet their position. Maybe I could compute `y` using the `bottom` helper, and x at 0.

I'll try to find a simpler workaround, thanks.

eric-burel:
Maybe you should merge the provided `data-grid` [here](https://github.com/STRML/react-grid-layout/blob/master/lib/utils.js#L613)  with the default item layout [here](https://github.com/STRML/react-grid-layout/blob/master/lib/utils.js#L623)? It's not uncommon to have only partial info about how you want the item, here I only know the size but I'd still like to have the default behaviour otherwise

STRML:
You could set `y` to `Infinity` and the compactor should simply put it at the bottom of the grid. See L44 of https://codesandbox.io/s/friendly-cookies-3gu2p, where I add a 26th item that is not in the passed layout.

eric-burel:
For the record, y = `Infinity` only works for compacted layouts. For non compact layouts you need to call `bottom(layout)` to get a correct behaviour without messing up the existing layout with new item.

You need to access the layout though, see https://github.com/STRML/react-grid-layout/issues/1133

Ideally, the new item should be compacted even in non compact mode, eg with a function like `bottom(layout, x, w)` that would compute the bottom but only for a restricted set of colums. But that's not doable yet and it's a different issue, `data-grid` is enough for me at this point.

Thanks for the insights!

linonetwo:
How to set initial width value?

I try to set initial width to 2 `<div key={key} data-grid={{ w: 2 }}>` but it says `Error: ReactGridLayout: ReactGridLayout.children[0].x must be a number!`

linonetwo:
I think this do the trick, if `isNewLayout` then apply the `defaultItemLayout`

```
export interface IMultiColumnProps {
  layouts: ReactGridLayout.Layouts;
  onChange: (newLayouts: ReactGridLayout.Layouts) => any;
  defaultItemLayout?: Partial<ReactGridLayout.Layout>;
}
export function MultiColumn(props: IMultiColumnProps): JSX.Element {
  const [currentBreakpoint, setCurrentBreakpoint] = useState('lg');
  const [allLayouts, setAllLayouts] = useState(props.layouts);
  const debouncedOnChange = useDebouncedCallback(props.onChange, [], 1000);
  const onLayoutChange = (layout: ReactGridLayout.Layout[]) => {
    const newAllLayouts = { ...allLayouts, [currentBreakpoint]: layout };
    /** is first created layout, we will apply some default value to it */
    const isNewLayout = allLayouts[currentBreakpoint] === undefined;
    if (isNewLayout) {
      newAllLayouts[currentBreakpoint] = newAllLayouts[currentBreakpoint].map((item) => ({
        ...item,
        ...(props.defaultItemLayout ?? {}),
      }));
    }
    setAllLayouts(newAllLayouts);
    debouncedOnChange(newAllLayouts);
  };
```

PR #1123, opened 2020-01-13
Add `shouldComponentUpdate` to major components.

STRML:
feat(render): Add shouldComponentUpdate to ReactGridLayout & GridItem.

This improves performance quite a bit in most cases, is implemented safely
to avoid breakage on equal arrays and objects.

To do so, while avoiding the maintenance burden of keeping a keylist
in sync with propTypes, we read propTypes using babel-plugin-preval
then construct a function to do the fastest possible comparison.

This is *a lot* faster than isEqual while avoiding unnecessary
comparison of e.g. `children`.

Supersedes #1032 - pinging @zaykaalexander.


STRML:
Resolved conflict with #1127 / #1128

n1ghtmare:
This looks ready to be merged. @STRML are there issues with it? I've tested it and everything seems ok. Need some help with it?

STRML:
Nope. Just concerned that it could cause the grid to miss updates in
certain conditions. I haven't found them though.

igorsha:
Hi! thanks for great  component. But after adding  shouldComponentUpdate shallow comparison,  when i'am using ResponsiveReactGridLayout it is is won't rendering at all! (obliviously why) i'am using dynamic count of elements with array inside body

{_.map(forms, (Form) => (
            <div key={Form.model.code} data-grid={{ x: 1, y: 0, w: 16, h: 12 }}>
                <Form />
            </div>
        ))}

thanks)

Issue #1107, opened 2019-12-18
Misplaced droppable marker when moving to RGL from another side.


jandufek:
Hello,
when I am "entering" a RGL with a Droppable element, initially it renders the marker at the top-left corner. It is not so obvious at the [demo page](https://strml.github.io/react-grid-layout/examples/15-drag-from-outside.html), because the source element is placed at the top-left as well, but if you enter it from a different side (e.g. your source element is in a panel at right side), it renders at the wrong place for a short moment and then it quickly moves to the correct one.
It is not so big problem, but it doesn't look very nice, which is bad, because the other animations look perfectly.

<img width="1440" alt="react-grid-layout-issue" src="https://user-images.githubusercontent.com/5647358/71072129-247d2a80-217e-11ea-9613-eef7df863336.png">

jandufek:
Does anybody have the same problem, please? If so I can look at the code and make a PR.

jandufek:
PR made here https://github.com/STRML/react-grid-layout/pull/1127

jandufek:
Fixed and merged to master

PR #1127, opened 2020-01-22
Moved calculation methods from GridItem to own util file and using it for dropping item.

jandufek:
I implemented changes for fixing this issue - https://github.com/STRML/react-grid-layout/issues/1107

The problem was, that the newly created dropping items were in `ReactGridLayout:onDragOver` set to coordinates `x:0, y:0` which caused the element jumpings.

So I decided to set the correct coordinates right from the beginning. It required the method for calculating the grid coordinates from pixels, which was already done in `GridItem:calcXY`. So only this method needed to be moved to a separate file,  but I thought, that it could be better to keep all calculation methods still together, that is why I moved all of them to new util file.


STRML:
Refactoring this actually works well with with #1123. Thanks for the contribution.

PR #1129, opened 2020-01-23
README - CSS instalation without node_modules path

dastasoft:
Changed Installation section of the README.md


STRML:
I don't want to assume everyone is using a webpack loader that supports that. For other applications, you might copy the files out.

Issue #1137, opened 2020-02-07
Flow fails due to incorrectly set defaultProps in `ResponsiveReactGridLayout`


mherzner:
`ResponsiveReactGridLayout`'s `Props` contain the `containerPadding` which are supposed to be: 
`{ [key: Breakpoint]: [number, number] } | [number, number]`, 
while the `containerPadding` within `defaultProps` is declared as:
`{ [key: Breakpoint]: null }`
thus causing `yarn flow` to fail

**steps to replicate:**
run `yarn flow` within the terminal

PR #1138, opened 2020-02-07
Fix flow error due to incorrectly set defaultProps

mherzner:
https://github.com/STRML/react-grid-layout/issues/1137

a simple fix for the issue i created.
sets the `containerPadding` within `defaultProps` to the expected value.


STRML:
Interestingly, I'm not getting this Flow error, but it *should* show up.

STRML:
Surfaced the error in https://github.com/STRML/react-grid-layout/commit/0e38ea00d59f84ab6bb3f2bda1ca6146e5ff15a6 and used that as an opportunity to fix up the props defs.

STRML:
This actually creates a behavioral difference as `null` `containerPadding` causes `<ReactGridLayout>` to use `margin`, so it would be `[10, 10]`, not `[0, 0]`. Reverting in 0.18.4.

Issue #1155, opened 2020-03-03
ReactGridLayout mutates props on widget resize


darekkay:
When adding a new item to the dashboard and resizing it immediately, props will be mutated [here](https://github.com/STRML/react-grid-layout/blob/master/lib/ReactGridLayout.jsx#L401). When using a library like [immer](https://github.com/immerjs/immer), the app will crash with the following exception:

```
Uncaught TypeError: Cannot assign to read only property 'w' of object '#<Object>'
  at ReactGridLayout.onResize (ReactGridLayout.js:614)
```

This is very similar to #1063 , but it's affecting another part of the code base (and is triggered by a different user action).

I will have a look into this and provide another fix.

darekkay:
I've created a PR with a fix: https://github.com/STRML/react-grid-layout/pull/1156

darekkay:
The issue isn't stale. I'll try to look back into my PR in the next days.

DWboutin:
Any news on this?

jaguardo:
Still an issue.

Issue #1063, opened 2019-11-01
ResponsiveReactGridLayout mutates props


darekkay:
After updating to `0.17.1`, my jest test is failing:

```
  const wrapper = mountConnected(<App />);
  ReactDOM.render(wrapper, div);
  ReactDOM.unmountComponentAtNode(div);
```

The error I get:

```
Cannot assign to read only property 'mobile' of object '#<Object>'
```

It is thrown here: https://github.com/STRML/react-grid-layout/blob/master/lib/ResponsiveReactGridLayout.jsx#L260

In addition, the application crashes, when changing the breakpoints.

The reason is that `layouts` is mutated, as it comes from `nextProps`. Instead, a new object should be created.
~As this particular code is not new, it might come from [this change](https://github.com/STRML/react-grid-layout/commit/3d4e863a0d4956b2d91542dfea4eaf52ff8e3a94#diff-632393c8029047f7a1ccf2e4bb3814e6R27).~

Mutating props usually doesn't break things (even though it is discouraged), but in my case, [immer](https://github.com/immerjs/immer) makes the props explicitly readonly, which in the end breaks.

The issue is reproducable in [my repository](https://github.com/darekkay/dashboard/blob/master/src/components/app/__tests__/index.test.tsx), after replacing `xit` with `it`.

I am happy to open a new PR if someone confirms my assumptions and the suggested solution.

P.S.: Thanks a lot for the latest updates!

darekkay:
This does not only let my test fail, but it also crashes the application when changing a breakpoint. I have provided and verified a fix in #1064 .

darekkay:
After a lot of debugging, I have finally found the core issue. It was not introduced with the latest react-grid-layout update, but with an update to `redux-starter-kit`. The change was introduced in [v0.9.0](https://github.com/reduxjs/redux-starter-kit/releases/tag/v0.9.0), most likely with the `immer` update. With this update, immer rightly assumes that the state/props are immutable. Hence, my fix is still relevant/required.

misha-panyushkin:
@darekkay @STRML any update on this?

I faced the relevant issue (as I supposed): 
```
Uncaught TypeError: Cannot assign to read only property 'w' of object '#<Object>'
```

Every time I add only one single item on the dashboard and trying to resize it crashes at this line:
```
at ReactGridLayout.onResize (ReactGridLayout.js:581)
```

yzvz:
I have the same issue @misha-panyushkin stated above:

[...]

PR #1156, opened 2020-03-03
Prevent prop mutation on item resize (#1155)

darekkay:
When adding a new item to the dashboard and resizing it immediately, props will be mutated [here](https://github.com/STRML/react-grid-layout/blob/master/lib/ReactGridLayout.jsx#L401). When using a library like [immer](https://github.com/immerjs/immer), the app will crash with the following exception:

```
Uncaught TypeError: Cannot assign to read only property 'w' of object '#<Object>'
  at ReactGridLayout.onResize (ReactGridLayout.js:614)
```

This is very similar to #1063 , but it's affecting another part of the code base (and is triggered by a different user action).

I've fixed the issue by cloning the active resize item in `onResize` (where the issue occurs) and updating that item in the layout in `onResizeEnd`. I had first tried cloning the entire layout in `onResize` (to prevent the temporary `activeResizeItem`), but this is highly ineffective, as it's executed on every mouse move.

I have confirmed the fix in my own project.

Fixes: #1155


STRML:
Thanks. I think a more comprehensive fix is needed here, to eliminate the state items `oldResizeItem`, and `oldDragItem`, and simply pluck them back out of the `oldLayout`.

STRML:
@darekkay I could really use a test that fails with the current code. You can use something like a deep Proxy freeze:

```js
// Deep freeze an object by using a Proxy.
// This is better than Object.freeze() as we can create coherent error messages
// and easily only deliver frozen subobjects when they are accessed. We can
// even add custom logic, like warning if you access a property that doesn't exist.
/* eslint-disable no-console */
export function deepFreeze<T>(inputObj: T): $ReadOnly<{|...T, toJSON: () => T|}> {
  // Our handler that rejects any change to the object and any nested objects inside it
  const deepFreezer = {};
  deepFreezer.get = function get<T: Object>(obj: T, prop: $Keys<T>, _receiver: Proxy<T>) {
    // Clone w/o Proxy
    if (prop === 'toJSON') return () => obj;
    // If dealing with nested object, nest the proxy untill it reaches the direct property of it's parent proxy
    if (typeof obj[prop] === 'object' && obj[prop] !== null) {
      return new Proxy(obj[prop], deepFreezer);
    }
    // If prop is directly accessible, just do the default operation
    else {
      if (!(prop in obj) && typeof prop !== 'symbol' && prop !== 'length' && prop !== '__esModule') {
        const stack = new Error().stack;
        logger.error(`Can not get unknown prop "${prop}" on frozen object. Stack: ${stack}`);
      }
      return obj[prop];
    }
  };
  deepFreezer.set = function set<T>(obj: T, prop: string, _val: any, _rec: Proxy<T>): boolean {
    const stack = new Error().stack;
    logger.error(`Can not set prop "${prop}" on frozen object. Stack: ${stack}`);
    // Note that we could return false, since returning false will create a TypeError,
    // but this could halt the whole program.
    return true;
  };
  // $FlowIgnore for all useful purposes the output type is T here
  return new Proxy(inputObj, deepFreezer);
}
```

DWboutin:
Any new development on this?

STRML:
Still waiting on a test so we can be sure this doesn't regress.

DWboutin:
Can I help for something?

STRML:
Sure. If you have a patch that adds a regression test here (i.e. fails on the old code, passes here), we can combine it with this and merge.

DWboutin:
Most of the work is done in the component (`onResize` and `onResizeStop`) and you just test the `utils`. I can write a test for the new `getLayoutItemIndex`, but without Enzyme, it will be difficult to test the component itself.

DWboutin:
I don't understand how you test the component in this repo since most of the work is done in the component `ReactGridLayout`.

STRML:
Yes, we already use enzyme in https://github.com/STRML/react-grid-layout/blob/master/test/spec/lifecycle-test.js.

DWboutin:
Good! I'll try something soon

DWboutin:
I just tried this fix, and it creates a box overlapping

DWboutin:
So, the fix provided don't work and give more bugs. I will not extend more on this because it's not a bug with react-grid-layout but with the data provided to the layout props.

I found that if you use Immer, you can use `setAutoFreeze(false)` to don't make your data readonly.

darekkay:
> So, the fix provided don't work and give more bugs. I will not extend more on this because it's not a bug with react-grid-layout but with the data provided to the layout props.

@DWboutin Thanks for checking this! Unfortunately, I don't have time to look into it again in the coming days. It is working fine for my use case (in production for weeks), but I agree that I might have missed some other use cases. However, it _is_ a bug within `react-grid-layout`, because React [props are considered immutable](https://reactjs.org/docs/components-and-props.html#props-are-read-only) (i.e. one should not mutate them):

> Whether you declare a component as a function or a class, it must never modify its own props. 

> I found that if you use Immer, you can use setAutoFreeze(false) to don't make your data readonly.

This is a workaround for this bug, but it just hides the core issue as mentioned above. It might also have other unwanted effects.

Could you maybe describe what's exactly broken with my PR? So I (or someone else) can build upon it instead of dismissing the whole issue.

christiankaseburg:
Anything I can do to help get this PR approved and merged? I'm seeing the same issue as well, and would like to help with this.

darekkay:
@christiankaseburg As @DWboutin mentioned, my PR seems to be introducing new bugs. So either those new bugs have to be fixed or a new PR with a different approach is required. In addition to that, some unit tests are required.

christiankaseburg:
@darekkay Do you have a work around for this temporarily? I am storing the layout in redux, and I am coming across issues with this error whenever trying to resize a grid-item.

garipov:
The same issue.
Tried to find a way how immutable layout gets into ReactGridLayout's state, but didn't have enough time.
However created a (sandbox)[https://codesandbox.io/s/react-grid-layout-error-8vxg0?file=/src/index.tsx] that reproduces this error. Probably it can be useful for someone who wants to help with this fix :)

Anwardo:
Experiencing the same issue. Would love if this was fixed. Setting `setAutoFreeze(false)` for me somehow only works after updating the layout at least once. The first resize action is still triggering
`Uncaught TypeError: Cannot assign to read only property 'w' of object '#<Object>' at ReactGridLayout.onResize (ReactGridLayout.js:614)`

jaguardo:
Is there any traction on this issue? I am unable to find a way to initiate `setAutoFreeze(false)` has that been deprecated?

QiuYeDx:
Has this been fixed in the latest version please?

STRML:

PR #1158, opened 2020-03-04
Fix example of memoizing children in README.md

draperunner:
The `count` variable comes from props, not as an argument to the useMemo callback.


STRML:
Thanks!

PR #1162, opened 2020-03-10
fix(drag): external drag must trigger activeDrag

### LINKED ISSUE #1152: Drag from outside is missing placeholder
Run /examples/15-drag-from-outside.html using version 0.18.x and note that the placeholder doesn't show up at all like it does in v0.17: https://strml.github.io/react-grid-layout/examples/15-drag-from-outside.html

paul-sachs:
Fixes #1152. When a drag event happens from an external element,
we should mimic the drag activity.

### DIFF
--- a/lib/ReactGridLayout.jsx
+++ b/lib/ReactGridLayout.jsx
@@ -661,7 +661,7 @@ export default class ReactGridLayout extends React.Component<Props, State> {
 
         if (!isEqual(prevCalculatedPosition, calculatedPosition)) {
           this.onDrag(
-            this.props.droppingItem.i,
+            droppingItem.i,
             calculatedPosition.x,
             calculatedPosition.y,
             {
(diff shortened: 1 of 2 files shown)


STRML:
Thanks for the contribution @paul-sachs. 

I'm not sure I understand how this broke between 0.17 and 0.18. It stands to reason that a patch to restore the old behavior shouldn't be this involved. Do you know?

Could you add a test for this? I've added some basic scaffolding (long overdue) for Enzyme tests in `master`. If you rebase upon it, it should be easy enough to add this.

STRML:
Bisect shows 1078b14582f424838947c8b0cf386eb6b45163b6 introduced the bug.

paul-sachs:
@STRML I'll take a closer look. I would also like to keep changes minimum. I'll look to rebase and integrate some tests, though drag events are notoriously complicated with nodejs dom implementations, so we'll see how that goes.

STRML:
Hey @paul-sachs. I got it working. PR to come shortly.

paul-sachs:
Closing in favour of #1164

Issue #1153, opened 2020-03-01
onDrop event is not invoking on ResponsiveReactGridLayout


hmudimi:
I am new to react-grid-layout, To know more about the project I have created a sample example with drag and drop. I have created a sandbox https://codesandbox.io/s/old-star-y3xb1. 

I have gone through the example in link  https://strml.github.io/react-grid-layout/examples/15-drag-from-outside.html, here drag and drop is working fine. While I am trying to use the same, the onDrop event is not firing (https://codesandbox.io/s/old-star-y3xb1).

ImAbdullahJan:
Update your packages to latest:
    "react": "16.12.0",
    "react-dom": "16.12.0",
    "react-grid-layout": "0.18.2",

and then test again.

hmudimi:
Thanks @ImAbdullahJan 
I have tried with the versions you suggested, Here onDrop event working fine but y value is always 0. Here it's considering only the x & y values of the entry-stage after that x & y values not changing. check URL https://codesandbox.io/s/old-star-y3xb1

So I tried react-grid-layout with 0.17.0 version, It's working as expected.

cs-miller:
yeah, there is an interesting problem where things work on 0.17.0, but you don't receive the event in onDrop, and in 0.18.2 you receive the event, but x and y are wrong and there is no placeholder element.

neilbryson:
It works properly on 0.17.1 too. 
I noticed on 0.18.2, onDrag and onDragStart do not get triggered when dragging an element from the outside. No placeholder and wrong coordinates, same with cs-miller's experience.

cedeber:
Fixed in 0.18.3.

hmudimi:
> Fixed in 0.18.3.

Yes, now it's working as expected.

PR #1164, opened 2020-03-11
Fix shouldComponentUpdate interfering with droppability

### LINKED ISSUE #1152: Drag from outside is missing placeholder

STRML:
Fixes #1152, #1153 and supersedes #1162

Please review @paul-sachs

### DIFF
--- a/__mocks__/react-grid-layout.js
+++ b/__mocks__/react-grid-layout.js
@@ -1,0 +1,2 @@
+// Used by jest as it requires some example files, which pull from 'react-grid-layout'
+module.exports = require('../index-dev');
--- a/examples/example-styles.css
+++ b/examples/example-styles.css
@@ -91,8 +91,9 @@ body {
 }
 .droppable-element {
   width: 150px;
-  background: #ddd;
+  text-align: center;
+  background: #fdd;
   border: 1px solid black;
-  margin-top: 10px;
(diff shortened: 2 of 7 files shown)


paul-sachs:
@STRML LGTM. Way less hacky than my solution by far. The only thing i noticed in the example, is that the `onLayoutChange1 event gets called once while dragging the element in. Not sure if that's a desired effect. This may have pre-existed though so might consider it a different issue.

STRML:
Yeah, there is a debate to be had about whether or not the `layout` coming from drag handlers actually should be updated, and whether or not `onLayoutChange` should fire only when the layout is finalized, or if it should fire every time something moves mid-drag or mid-resize. 

That said, I don't want to change any of those assumptions right yet. It would be great to have more tests asserting current callback behavior so we can be sure we don't break userland, then iterate on it in major versions or via new callbacks as-needed.

paul-sachs:
@STRML yeah that makes sense. It's an easy enough thing to workaround really, just filter out the `droppingItem` objects whenever we want to save the layouts.

PR #1167, opened 2020-03-16
docs: add null to compactType options in README

ShadyMohammed:


STRML:
That's what the `?` signifies ("maybe" type).

PR #1169, opened 2020-03-18
Consistency: make onDrop callback more consistent

cedeber:
Hello,

I need to have access to the layout on the onDrop callback. In other words, this PR doesn't solve any bug. But this callback still looks dirty compared to other callbacks, so I also try to make it more consistent with the rest of the app.

We can still improve that a bit, but I already create a PR so that we can talk about it as it is a major change to the way the callback is used.

Thanks


STRML:
Thanks, this is a good idea. I'm doing a semver-major release (1.0.0) and this will make it in.

Issue #1170, opened 2020-03-19
Missing documentation for maxRows


SanjiKir:
During development I've met an issue that I was able to drop element from parent layout anywhere I want vertically, thus prolonging the Y-axis infinitely. Once I started digging into your code I found out that there is actually a prop which limits the number of vertical rows which is not documented

PR #1171, opened 2020-03-19
Update README.md

SanjiKir:
https://github.com/STRML/react-grid-layout/issues/1170


STRML:
See https://github.com/STRML/react-grid-layout/blob/83251e5e682abfa3252ff89d4bacf47fdc1f4270/lib/ReactGridLayoutPropTypes.js#L110-L115. I'd like to fix `maxRows` before exposing it.

PR #1176, opened 2020-03-22
Add Ref to Grid

paul-sachs:
Add reference prop so we can get the internal div element. Useful for measurement and scroll control.

We could use React.forwardRef but that does a few things:
- limits version of compatible react
- breaking API change (ref would now be the div, vs the class object)

### DIFF
--- a/README.md
+++ b/README.md
@@ -355,7 +355,10 @@ onResize: ItemCallback,
 // Calls when resize is complete.
 onResizeStop: ItemCallback,
 // Calls when some element has been dropped
-onDrop: (elemParams: { x: number, y: number, w: number, h: number, e: Event }) => void
+onDrop: (elemParams: { x: number, y: number, w: number, h: number, e: Event }) => void,
+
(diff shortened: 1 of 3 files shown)


STRML:
That's pretty useful, especially since `ReactDOM.findDOMNode()` is discouraged now. 

It looks like this is pretty unsurprising / used elsewhere (https://github.com/atlassian/react-beautiful-dnd/blob/master/docs/guides/using-inner-ref.md), so I'll merge.

STRML:
Thanks for the contribution!

paul-sachs:
@ben-elsen not sure what you mean by implement. To use the new prop:

```jsx
const ref = useRef();
return (
  <GridLayout innerRef={ref} />
);
```

This change didn't remove any references to findDOMNode within the source. More work would need to be done to remove the usage of findDOMNode.

ben-elsen:
@paul-sachs I misunderstood this feature.. I thought it was intended to fix the warning. Sorry :-)

PR #1178, opened 2020-03-26
Resize placeholder should not capture pointer events

bripkens:
# Why

Currently the resizing placeholder captures pointer events, e.g. while resizing the
cursor bounces back and forth between the placeholder and the actually
resized item.

This in turn can result into undesired side-effects/flickering when users modify
the resizing edge via CSS, e.g. when adding a hover effect to the resizing
edge.

# What

Disable pointer events. This way the placeholder never captures the pointer
events (neither for JS nor CSS).

### DIFF
--- a/css/styles.css
+++ b/css/styles.css
@@ -34,6 +34,7 @@
   -ms-user-select: none;
   -o-user-select: none;
   user-select: none;
+  pointer-events: none;
 }
 
 .react-grid-item > .react-resizable-handle {


bripkens:
I still believe this is a valid change for this project and the PR should not be considered "stale" :)

STRML:
Could you show an example that demonstrates this problem?

bripkens:
Hey @STRML,

this happens when shrinking an item slightly below its current size. In this moment the placeholder is larger than the resized item and therefore captures the pointer event.
This effect can best be seen in a quick video. Here you go:

https://www.youtube.com/watch?v=ZOWQqoTyBMA&feature=youtu.be

Issue #1100, opened 2019-12-10
How to achieve horizontal + Vertical compacting together?


nitesh619:
Can we have both horizontal and vertical compacting together?

Hari-Shankar-3296:
We can achieve that but the grid packing will be very weird.
Set the props compactType = "horizontal" and verticalCompact = true
FYI Check all possible arrangements.

nitesh619:
@Hari-Shankar-3296 Yes, I tried setting the verticalCompact prop and it is super weird.

But my goal is to always move tiles from right to left. If i drag a tile from the top row to the bottom row. Then the spot vacated by this tile should be occupied by tile to its right and the tile from the beginning of next row should come up and so on. Like text in paragraph, if we delete letters then next line character will come up.

Hari-Shankar-3296:
@nitesh619 I got your point now. The thing you are expecting ( even I need the same behaviour ) is not possible with react-grid-layout currently. But there are other library in which you can achieve this.

Here is the link. 
https://ourcodeworld.com/articles/read/663/top-5-best-draggable-droppable-and-resizable-grid-layout-designer-jquery-and-javascript-plugins
and up to me Muuri is the best.
```Note :```
```
In Murri we can't resize grid items ( Can be done by adding React-Resizable ).
Some libraries are done using jQuery.
Up to my knowledge Muuri doesn't use jQuery
```
If you find any other good libraries, Please do let me know. :+1:

nitesh619:
Thanks @Hari-Shankar-3296, Murri is really good. But I am not sure how it will work with react and also I can't afford to revert my changes and start with murri, I guess I have passed that stage. So I have decided to modify react-grid-layout codebase for the behavior I need. Wish me luck, looks like a tough task ahead. I will keep you posted.

JoshRozin:
@nitesh619 Hey, Did you end up getting that working? I've seen react-wrappers for muuri, not sure if they're still up to date.

nitesh619:
Hey,

Yes, actually I was able achieve this behavior by easily overriding few util methods of the grid.

amrsalama:
@nitesh619 May you share your solution with us ?

IvoBeneStudio:
@nitesh619 Little reminder if you would please share your solution. Much needed here as well :)

nitesh619:
### How to do horizontal + vertical compaction combined,  also maintain ordering while dragging tiles on the grid
[...]

PR #1188, opened 2020-04-15
Implement dragging in "wrap" mode

JohnThomson:
Handles moving conflicting elements in a way appropriate for a sequence of elements arranged in a wrapped-paragraph style.


JohnThomson:
This uses code adapted from a comment by nitesh619 in https://github.com/STRML/react-grid-layout/issues/1100. However, rather than just a one-off patch, I have tried to make a full PR with unit tests and a new compactType value to invoke this behavior.

JohnThomson:
Requires a change to the corresponding @types project to permit the new "wrap" value for compactTypes. I'm not sure whether that will happen automatically when this is merged or whether it requires a separate PR.

STRML:
While I don't dislike or deny this use case, it is too esoteric to be included in the main library. I would rather that compaction be passable as a function that does the compaction, and then users can produce their own compactors at will without modifying the main library.

kapil-kharade:
Any updates on this PR?

JohnThomson:
I'm sorry, I don't know when or whether I will have time to rework this along the lines STRML wants.

OZZlE:
I think this 'wrap' case is what 99% of people would want rather than the 'vertical' or 'horizontal' but we cannot include it in code?

We're using it to simply re-order things in a grid, I haven't been able to replicate it but got reports that the current 'horizontal' mode sometimes sends items on the far right of the grid all the way to the bottom of the page.. which is super annoying with hundreds of items.

We have a grid like this:

    [1][2][3][4]
    [5][6][7][8]

and you might want to drag number 6 to number 2's position, then 2, 3 and 4 moves to the right which is expected but number 4 goes to position 8 or randomly 999.. which is not wanted at all probably by anyone.. (we don't have any empty slots just re-ordering can be done)

The releases doesn't have any notes so I cannot see if it has been fixed in the latest version and if it's worth just to update. If number 4 would just go to position 8 it would be fine but when it ends up in the bottom of the grid it's a big issue...

STRML:
See the changelog.

farion:
I also think, that the wrap behavior ist what people usually expect. Everything else is considered as "wrong" on the first sight. So my question is, what needs to be done to get this merged or what is the current recommendation to get this behavior with the current release?

nelsieborja:
> I also think, that the wrap behavior ist what people usually expect. Everything else is considered as "wrong" on the first sight. So my question is, what needs to be done to get this merged or what is the current recommendation to get this behavior with the current release?

Exactly. We also need this feature so have tried to find an alternative and came across [Gridstack](https://gridstackjs.com/demo/two.html) although it doesn't have this horizontal compacting, it has a `compact` method (compacts on both) which you can call manually. I personally prefer it now coz of its flexible resizing feature.

[Here's](https://vibvy.csb.app/) my WIP playground for it or you can just check the library's demo, ofc ;)

kasmoto:
A prayer that someone incorporates this via method prescribed.. 
Something a lot of people are looking for, not only within this library but others too.

rbarkhouse:
I would also like to +1 the request for a "wrap" compact type, it is exactly what we are looking for and is our only complaint with this otherwise great library.

AntSan813:
+1

kevinschaffter:
+1

ArianKrasniqi:
+1

Saransh27:
Hi @STRML I would like to request you or some other maintainer to please review and merge this PR? I also need this feature, such an amazing library, only the "wrap" compact type  is missing.

fmiskuly-eastman:
+1 to wrap functionality

STRML:
Continued in #1773.

Issue #1186, opened 2020-04-13
Uncaught TypeError: Super expression must either be null or a function


igor9silva:
Just leaving it here as a reference for someone else.

The following error started happening on my web app a couple hours ago:
```
Uncaught TypeError: Super expression must either be null or a function
```

The only hint was pointing to https://github.com/STRML/react-draggable.

What happened is that `grid-layout@v0.17.1` depends on `react-draggable@^4.0.0`, which deployed `v4.3.1` ~6 hours ago.

I'm not sure if `react-draggable@4.3.1` is broken or not backward-compatible.

My solution was to add into my `package.json`:
```
"resolutions": {
    "react-grid-layout/**/react-draggable": "4.2.0"
},
``` 

That way `yarn` (as per https://classic.yarnpkg.com/en/docs/selective-version-resolutions) could do the dirty work for me.

There is also `npm-shrinkwrap` which might be helpful if you're not using `yarn`.

I'd recommend updating the depency from `^4.0.0` (minor) to `4.2.x` (patch, and in a more readable form).

immanuelfodor:
Feeling the same heat :(

![image](https://user-images.githubusercontent.com/21174107/79341557-4ae48400-7f2c-11ea-87b9-243682416084.png)

immanuelfodor:
It was all good until yesterday where our Verdaccio npm proxy cached a new version, so we're trying now to implement your resolution fix, maybe it helps us, too:

![image](https://user-images.githubusercontent.com/21174107/79341979-db22c900-7f2c-11ea-9a8e-14c2cd7fb8f7.png)

immanuelfodor:
For npm, we made the following changes, now it works again:

**package-lock.json**
```diff
-      "version": "4.3.1",
-      "resolved": "https://registry.npmjs.org/react-draggable/-/react-draggable-4.3.1.tgz",
-      "integrity": "sha512-m8QeV+eIi7LhD5mXoLqDzLbokc6Ncwa0T34fF6uJzWSs4vc4fdZI/XGqHYoEn91T8S6qO+BSXslONh7Jz9VPQQ==",
+      "version": "4.2.0",
+      "resolved": "https://registry.npmjs.org/react-draggable/-/react-draggable-4.2.0.tgz",
[...]

PR #1199, opened 2020-04-25
config(dependency): upgrade react-draggable

solo474:
RGL is throwing an error in production mode of create-react app. likely due to uglify or tarsier optimisation. 
ref: #1186

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "classnames": "2.x",
     "lodash.isequal": "^4.0.0",
     "prop-types": "^15.0.0",
-    "react-draggable": "^4.0.0",
+    "react-draggable": "^4.3.1",
     "react-resizable": "^1.10.0"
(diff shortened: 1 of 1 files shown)


igor9silva:
It has a `--frozen-lockfile` flag on yarn build. Not sure why.

@STRML can we update it?

igor9silva:
@solo474 updating to `^4.3.1` won't fix the problem as we need `4.2.0`.

I'd recommend using `4.2.x` to upgrade only minor versions and be more readable.

**Source:** https://docs.npmjs.com/about-semantic-versioning

STRML:
This PR won't fix any issue - incrementing the caret version range isn't going to do anything meaningful, and 4.3.1 is apparently the broken version anyway. Can anybody reproduce this adequately so we can simply fix it?

Issue #1218, opened 2020-06-01
"onLayoutChange" callback is not called for an unusual case


levvsha:
I faced with a case when `onLayoutChange` callback is not called, although the layout was actually changed. In our application we have large (with a lot of cells) tables as children for `ReactGridLayout` component. I have created the simplified [example](https://jsfiddle.net/jeLgb1o4/) so you can check that if you change the width of the table quickly, the `onLayoutChange` callback is not triggered (see the attached gif).

After debugging I realized that the cause of the problem is in [this line](https://github.com/STRML/react-grid-layout/blob/master/lib/ReactGridLayout.jsx#L364). `this.state.layout` is an object so here it saves as a reference and the reference will be in[ this variable](https://github.com/STRML/react-grid-layout/blob/master/lib/ReactGridLayout.jsx#L373) too. Setting `w` and `h` property for `l` [here](https://github.com/STRML/react-grid-layout/blob/master/lib/ReactGridLayout.jsx#L402) updates the values for the appropriate element in `this.state.oldLayout`. As a result [this](https://github.com/STRML/react-grid-layout/blob/master/lib/ReactGridLayout.jsx#L352) condition returns `false` (`oldLayout` and `newLayout` would be equal). After the second call of the `onResize` method everything would be OK because the `compact` method [here](https://github.com/STRML/react-grid-layout/blob/master/lib/ReactGridLayout.jsx#L420) will return copy of `layout` (not a reference) so this problem does not affect 99.99% cases (`onResize` is triggering many time during resize) but in my case the large table with a lot of cell forces a browser (or maybe react) to enable some optimization technics and in my case the `onResize` method sometimes is called only once.

Surely, I have a problem on my side too - it is not OK to have so large table without virtualizing, but storing by reference should be avoided. I have created [the PR](https://github.com/STRML/react-grid-layout/pull/1219), please take a look.

![resize-problem](https://user-images.githubusercontent.com/19476358/83417346-a0b4b300-a42a-11ea-8182-281f7e6055ca.gif)

STRML:
Hey, thanks for this. I think we need a bit more time to think about this, because there is a somewhat unstructured approach to when layout items are cloned, and when they are not. Ideally, items to-be-modified are never actually modified, but cloned and placed into a new array, which speeds up equality checks, but slows down everything else and generates a lot of garbage if not handled well.

[...]

PR #1219, opened 2020-06-01
Clone layout array before storing to oldLayout

levvsha:
Clone `this.state.layout` before updating `this.state.oldLayout` property.

Currently `this.state.layout` value is storing by reference to `this.state.oldLayout` as a result some incorrect behavior can happen (see the reference issue).

Reference issue - https://github.com/STRML/react-grid-layout/issues/1218


STRML:

PR #1233, opened 2020-06-24
Fix GrildItem cannot  be used

coralf:


STRML:
I don't understand the reason for this PR. Could you explain?

coralf:
> I don't understand the reason for this PR. Could you explain?
### The problem recurrence steps are as follows
1. Set the element of containing the GridLayout display:none.
2. Change browser window size.
3. Display the element.
These steps can lead to the node.offsetWidth is undefined in the onWindowResize.
` onWindowResize = () => {
      if (!this.mounted) return;
      const node = ReactDOM.findDOMNode(this);
      if (node instanceof HTMLElement && node.offsetWidth)
        this.setState({ width: node.offsetWidth });
    };
`

STRML:
This patch doesn't seem to fix the issue. It couldn't, in fact, as if `node.offsetWidth is undefined` is an error you actually see, it is because `node` is undefined, which is not possible. 

We simply can't measure the element while `display: none` is set. The only way to detect if it's unset is to use a MutationObserver, which you could do, but really should be done on the application side (fire a 'resize' event) and not in the library.

PR #1246, opened 2020-07-17
Fix #1243

### LINKED ISSUE #1243: Grid Items are stuck using css percentages on first render.
When the grid is first rendered, the grid's `mounted` state is `false`. When grid items are rendered in this state, they use `usePercentages`. For some reason, after the grid has mounted, the grid items don't get re-rendered and so they are stuck in "percentages" mode until a user interacts with the grid.

This was not the case in 0.16.x

Sandbox: https://codesandbox.io/s/react-grid-layout-stuck-percentages-j75gm

You'll need to use the in-app browser refresh button to see the issue.

Screen Shot 2020-07-15 at 6 12 44 PM

rhgb:
#1243 This issue is caused by:

1. `cssTransforms` props of `GridItem` is determined by combining two things in `ReactGridLayout`: the property `cssTransforms`, and the state `mounted`.
2. The `mounted` state is initially false, causing `GridItem` using offset and percentage for positioning initially, instead of transform.
3. The `mounted` state is set to true in `componentDidMount`. But because `shouldComponentUpdate` method doesn't check it, the `GridItem` rendered won't change, until another update is triggered.

This PR fixes this issue. I've checked that it works in my application.

### DIFF
--- a/lib/ReactGridLayout.jsx
+++ b/lib/ReactGridLayout.jsx
@@ -203,6 +203,7 @@ export default class ReactGridLayout extends React.Component<Props, State> {
       this.props.children !== nextProps.children ||
       !fastRGLPropsEqual(this.props, nextProps, isEqual) ||
       this.state.activeDrag !== nextState.activeDrag ||
+      this.state.mounted !== nextState.mounted ||
       this.state.droppingPosition !== nextState.droppingPosition
     );
   }


STRML:
Thanks!

mario-jerkovic:
Is there any eta for this fix :)

STRML:
Yes, it's in `1.0.0`.

PR #1269, opened 2020-08-24
Fix for the incorrect drag behavior when a parent element has a trans…

tculig:
Noticed this kind of behavior happening again: https://github.com/STRML/react-grid-layout/issues/1046
Here is also a live version: https://5wy3rz5z1x.csb.app/

Investigations led to the conclusion that both the domFns.js of react-draggable, and GridItem.js of react-grid-layout are, are scaling the deltaX and deltaY when dragging. Thus the scaling is applied twice.

I've deleted the instance that happens in the GridItem.js and now its working fine for me.

Hope this helps.

### DIFF
--- a/lib/GridItem.jsx
+++ b/lib/GridItem.jsx
@@ -425,8 +425,6 @@ export default class GridItem extends React.Component<Props, State> {
   onDrag = (e: Event, { node, deltaX, deltaY }: ReactDraggableCallbackData) => {
     const { onDrag, transformScale } = this.props;
     if (!onDrag) return;
-    deltaX /= transformScale;
-    deltaY /= transformScale;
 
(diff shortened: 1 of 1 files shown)


joeyzhousherwin:
Please take a look at this, this is currently preventing me from doing what I need with cssTransform

STRML:
Thanks, you are correct, I have tested this locally. Would you mind adding a test for this behavior?

tculig:
Eh, I knew just enough programming to fix the issue, but I don't know how to write a test. :/

tculig:
This bug is still in the code...everytime I update the lib I have to manually change it. I don't have time to learn how to write tests. Any chance someone else could do it??

PR #1284, opened 2020-09-10
fix: remove dist folder from npm ignore

jmfrancois:
Today we can t use the lib throw any CDN, like unpkg.

This change is just to let people use the dist version on codepen, or any other external service.

Note it has never been published :
https://unpkg.com/browse/react-grid-layout@0.10.0/
https://unpkg.com/browse/react-grid-layout@1.0.0/

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,6 +1,5 @@
 test/
 examples/
-dist/
 yarn.lock
 .github
 *.config.js


STRML:
Thanks, this is a good idea. I am considering actually removing them from Git as they really should only belong in the published package. There should be no good reason to install this package from git *and* require a UMD build.

Issue #593, opened 2017-07-20
`onResizeStop` not firing `onLayoutChange` reliably


thomb:
- [x] Bug
- [ ] Feature Request
- [ ] Question

I've run into a situation where `onResizeStop` doesn't trigger `onLayoutChange` all the time.

The `layout` as returned by the `onResizeStop` handler reflects the changes to the layout, but `oldLayout` and `newLayout` are equal. meaning on that `onLayoutMaybeChanged` never fires `onLayoutChange`.

Apologies for the lack of a test case, the `webpackBin` link in the ticket wasn't working.

al-fyodorov:
Hey @thomb, you can send steps for reproduce this bug?

STRML:
#1289 needs more explanation, but I can see where there may be an issue here if `state.layout` is mutated.

PR #1289, opened 2020-09-15
fix `onResizeStop` not firing `onLayoutChange`

xcqwan:

example or test case.

ref issue: https://github.com/STRML/react-grid-layout/issues/593


xcqwan:
@daynin have a look, thx!

STRML:
Thanks for this. We can save time by only cloning the item that we mean to modify. See the commit above.

PR #1296, opened 2020-09-21
Protect against children that may be null, this happens with conditio…

jeffras:
…nal rendering https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator

example or test case.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -127,8 +127,8 @@ export function cloneLayoutItem(layoutItem: LayoutItem): LayoutItem {
  */
 export function childrenEqual(a: ReactChildren, b: ReactChildren): boolean {
   return isEqual(
-    React.Children.map(a, c => c.key),
-    React.Children.map(b, c => c.key)
+    React.Children.map(a, c => c ? c.key : ""),
+    React.Children.map(b, c => c ? c.key : "")
   );
 }
 
@@ -647,6 +647,9 @@ export function synchronizeLayoutWithChildren(
   // Generate one layout item per child.
   const layout: LayoutItem[] = [];
   React.Children.forEach(children, (child: ReactElement<any>, i: number) => {
+    if (!child || !child.key)
+      return
+
     // Don't overwrite if it already exists.
     const exists = getLayoutItem(initialLayout, String(child.key));
     if (exists) {


jeffras:
Added issue 1297 for this PR

lib/utils.js
```diff
@@ -647,6 +647,9 @@ export function synchronizeLayoutWithChildren(
   // Generate one layout item per child.
   const layout: LayoutItem[] = [];
   React.Children.forEach(children, (child: ReactElement<any>, i: number) => {
+    if (!child || !child.key)
```
STRML, line 650: This will bail out if `child.key === 0`

STRML:
I haven't seen this come up before, do you have an example of where this is causing problems?

jeffras:
Here is the code that reproduces the issue. I believe using the inline condition returns a null child on the false case

`import React from "react";
import { WidthProvider, Responsive } from "react-grid-layout";

const ResponsiveReactGridLayout = WidthProvider(Responsive);
const originalLayouts = getFromLS("layouts") || {};

/**
 * This layout demonstrates how to sync multiple responsive layouts to localstorage.
 */
export default class ResponsiveLocalStorageLayout extends React.PureComponent {
  constructor(props) {
    super(props);

    this.state = {
      layouts: JSON.parse(JSON.stringify(originalLayouts))
    };
  }

  static get defaultProps() {
    return {
      className: "layout",
      cols: { lg: 12, md: 10, sm: 6, xs: 4, xxs: 2 },
      rowHeight: 30
    };
  }

  resetLayout() {
    this.setState({ layouts: {} });
  }

  onLayoutChange(layout, layouts) {
    saveToLS("layouts", layouts);
    this.setState({ layouts });
  }

  render() {
    const test = false; //NOTE: true works fine, but false seems to return a null child which crashes

    return (
      <div>
        <button onClick={() => this.resetLayout()}>Reset Layout</button>
        <ResponsiveReactGridLayout
          className="layout"
          cols={{ lg: 12, md: 10, sm: 6, xs: 4, xxs: 2 }}
          rowHeight={30}
          layouts={this.state.layouts}
          onLayoutChange={(layout, layouts) =>
            this.onLayoutChange(layout, layouts)
          }
        >
        {test &&
          <div key="1" data-grid={{ w: 2, h: 3, x: 0, y: 0, minW: 2, minH: 3 }}>
            <span className="text">1</span>
          </div>
          }
          <div key="2" data-grid={{ w: 2, h: 3, x: 2, y: 0, minW: 2, minH: 3 }}>
            <span className="text">2</span>
          </div>
          <div key="3" data-grid={{ w: 2, h: 3, x: 4, y: 0, minW: 2, minH: 3 }}>
            <span className="text">3</span>
          </div>
          <div key="4" data-grid={{ w: 2, h: 3, x: 6, y: 0, minW: 2, minH: 3 }}>
            <span className="text">4</span>
          </div>
          <div key="5" data-grid={{ w: 2, h: 3, x: 8, y: 0, minW: 2, minH: 3 }}>
            <span className="text">5</span>
          </div>
        </ResponsiveReactGridLayout>
      </div>
    );
  }
}

function getFromLS(key) {
  let ls = {};
  if (global.localStorage) {
    try {
      ls = JSON.parse(global.localStorage.getItem("rgl-8")) || {};
    } catch (e) {
      /*Ignore*/
    }
  }
  return ls[key];
}

function saveToLS(key, value) {
  if (global.localStorage) {
    global.localStorage.setItem(
      "rgl-8",
      JSON.stringify({
        [key]: value
      })
    );
  }
}`

jeffras:
And the error
image

lib/utils.js
```diff
@@ -647,6 +647,9 @@ export function synchronizeLayoutWithChildren(
   // Generate one layout item per child.
   const layout: LayoutItem[] = [];
   React.Children.forEach(children, (child: ReactElement<any>, i: number) => {
+    if (child === null || child.key === null)
```
STRML, line 650: Might be undefined, so I would do `if (child == null || child.key == null)`

lib/utils.js
```diff
@@ -647,6 +647,9 @@ export function synchronizeLayoutWithChildren(
...
+      return
```
STRML, line 651:
```suggestion
      return;
```

lib/utils.js
```diff
@@ -721,10 +724,10 @@ export function validateLayout(
     if (item.static !== undefined && typeof item.static !== "boolean") {
       throw new Error(
         "ReactGridLayout: " +
-          contextName +
-          "[" +
-          i +
-          "].static must be a boolean!"
+        contextName +
```
STRML, line 727: Please refrain from stylistic updates in your commits.

PierrickI3:
Thanks! Issue is fixed now.

PR #1298, opened 2020-09-22
Add isBoundedX and isBoundedY

LorenzHenk:
> *Please reference an open issue. If one has not been created, please create one along with a failing
example or test case.*

I'm not sure about the process here. In the issue, it says "do only use issues for bug tracking".

## Description

Bound x axis is cool, but we need the y axis to be free to automatically resize the grid.

Thus the proposal for the flags `isBoundedX` and `isBoundedY` to disable the bounding box for a specific axis.

If you are fine with breaking changes, we could also remove `isBounded`.


STRML:
Closing: This feature/fix is now included in the v2 TypeScript rewrite. The functionality has been incorporated into the core architecture. Thank you for the contribution!

Issue #1276, opened 2020-09-04
Resizable handlers incorrectly resizing 


LorenzHenk:
![resize](https://user-images.githubusercontent.com/10603631/92234041-a0a4d300-eeb1-11ea-8fa7-31e861a00749.gif)

As you can see in the gif above, resizing with handles on the top or right do not work as expected. They should also edit the `x` and `y` properties of the item, not just the `width` and `height`.

See the new example (17, resizable handles). Seems like the following link (https://strml.github.io/react-grid-layout/examples/17-resizable-handles.html) is not working, I believe you have to upload the built examples.

MarkyDeParky:
I'm experiencing the same issue as @LorenzHenk . I've set the resizeHandles to "sw" and "se" and in my case "sw" is not behaving like I expect it to. I would expect it to 'expand' to the left when I move my mouse to the left. Also, the mouse cursor on Mac isn't correct (and different from what @LorenzHenk shows in his animation). On Mac it's the same arrow pointing to bottom right as with the SE handle.

![Sep-22-2020 12-09-34](https://user-images.githubusercontent.com/16799069/93869935-cf51e480-fccc-11ea-999e-ba36105cca3c.gif)

Another thing that would be nice, but maybe this is more of a feature request, is that handles only appear if expanding is possible in that direction. So "SE" would only appear if the item isn't already 'docked' to the far right, and "SW" would only appear if the item isn't docked to the far left. 

![Sep-22-2020 12-11-35](https://user-images.githubusercontent.com/16799069/93870328-5b640c00-fccd-11ea-92b0-844454325871.gif)

LorenzHenk:
> Also, the mouse cursor on Mac isn't correct (and different from what @LorenzHenk shows in his animation). On Mac it's the same arrow pointing to bottom right as with the SE handle.

I do have this problem as well. It doesn't appear in my recording, because for the recording I modified an official example. The examples have different styles applied. In [the example styles](https://github.com/STRML/react-grid-layout/blob/master/examples/resizable-styles.css), the cursor is set correctly.
In the [actual styles](https://github.com/STRML/react-grid-layout/blob/master/css/styles.css#L45), the other cursors are not styled differently.

I'll create a PR for the cursor styles problem.

> Another thing that would be nice, but maybe this is more of a feature request, is that handles only appear if expanding is possible in that direction. So "SE" would only appear if the item isn't already 'docked' to the far right, and "SW" would only appear if the item isn't docked to the far left.

[...]

PR #1299, opened 2020-09-22
Include resizable handles styles

LorenzHenk:
Issue comment mentioning the problem: https://github.com/STRML/react-grid-layout/issues/1276#issuecomment-696632566

> ... it's the same arrow pointing to bottom right as with the SE handle.

This PR fixes the incorrect cursor styles.


STRML:
Thanks! #1276 is a larger issue that will require some more thought.

PR #1303, opened 2020-09-30
Pass handle option to react-resizable

typeetfunc:
Problem: we can't pass custom resize handle to react-resizable
Solution: I add prop 'resizeHandle' to ReactGridLayout


daynin (approved):

STRML:
Clean!!!! Thank you.

typeetfunc:
@STRML Can you release a new  package version with this changes?

TikTikTik21:
The resizeHandle props is still not working

BboyAwey:
The package on npm still not work with this resizeHandle prop. Weird~~~

AaronPowell96:
> The package on npm still not work with this resizeHandle prop. Weird~~~

Works for me mate, pass a function and then return an element I use a switch to check if the axis is e s se etc.

Why isnt this also added for dragHandle?

Issue #1308, opened 2020-10-05
Adding static items to a layout can cause non-static items to be placed in wrong location


jenniferarnesen:
The use case is adding "page breaks" (static items) to a dashboard built with react-grid-layout when user wants to print the dashboard.

Here is a codepen demonstrating the problem. Click on the "Add page breaks" button to see that item `F` incorrectly gets moved up in the layout:  https://codepen.io/jjarnesen/pen/JjKPGPm. (apologies, I didn't see until later that RGL requested CodeSandbox to be used)

I have created a patch to fix the problem in our web application, and submitted a PR to RGL here: https://github.com/STRML/react-grid-layout/pull/1309. You might decide to implement a fix differently.

PR #1309, opened 2020-10-05
fix: do not optimize item positioning if there are static items in the layout

jenniferarnesen:
If there are statics in the layout, then all items could potentially need moving, so you can't break out of the item-moving code early.

Fixes: https://github.com/STRML/react-grid-layout/issues/1308


jenniferarnesen:
Keeping it alive

PR #1310, opened 2020-10-05
onDragOver firefox issue

romeo-str:
line 588: e.nativeEvent.target.className.indexOf(layoutClassName)
give "TypeError: e.nativeEvent.target.className.indexOf is not a function"

replaced by:
!e.nativeEvent.target.classList.contains(layoutClassName)


stephsprinkle:
@STRML Would it be possible to get this fixed merged in? Our project is also affected by this error.

jacekpolak:
@STRML could you please share when this fix could be released to 1.3.0 ?

PR #1092, opened 2019-11-28
Replace ReactDOM.findDOMNode with React.createRef

Ginioo:
Hi,

This PR is trying to solve this issue: https://github.com/STRML/react-grid-layout/issues/1091

PR #1318, opened 2020-10-21
Fix/replace finddomnode noconflicts

westandy:
This is my attempt to clean up https://github.com/STRML/react-grid-layout/pull/1092 to get the Ref changes.  

Thank you @STRML and others for any feedback.


westandy:
Note: I do not have write access.  Can someone else merge?

Marvimoto:
Hey, I just tested it and it does not seem to work. In dist/react-grid-layout.min.js there are still findDOMNodes.

lib/components/WidthProvider.jsx
```diff
@@ -77,7 +83,7 @@ export default function WidthProvider<Config>(
     onWindowResize = () => {
       if (!this.mounted) return;
       // eslint-disable-next-line react/no-find-dom-node
-      const node = ReactDOM.findDOMNode(this); // Flow casts this to Text | Element
+      const node = this.currentNode.current; // Flow casts this to Text | Element
```
STRML, line 86: I don't think this will work unless we forward the underlying actual DOM ref from the underlying Grid Layout.  This is going to be a `React.Ref<typeof ReactGridLayout>`, not `React.Ref<HTMLElement>`.

STRML:
Concept ACK here, but please remove the massive style changes and please test with the examples.

STRML:
Fixed and released in `1.2.3` via #1425.

PR #1323, opened 2020-10-30
fix: Modify Drag values according to containerPadding

hywlss9:
If you drag the item with containerPadding, it looks as if it is moved more than the actual location.


STRML:
Can't merge with all these project changes - please remove the changes to `dist`, `.gitignore`, `build`, etc., then we can review it.

hywlss9:
@STRML i remove dist, .gitignore, build ~ I'm sorry I'm late.

mvrozanti:
@STRML hi

is there anything holding back this PR?

STRML:
Wow, this is actually a big fix - if `containerPadding` is at all significant, the calculations were really wrong.

I kind of regret this even being an option. It would be better if one used an outer container with padding or had margin on the grid layout. But here we are, and thank you for the fix!

STRML:
`1.4.4`

mvrozanti:
Nice!

hywlss9:
Thank you!
Hope you enjoy using this :)

wanpan11:
ref: https://github.com/react-grid-layout/react-grid-layout/issues/2101

Issue #1324, opened 2020-11-03
Layout Incorrect while Drag-Drop New Element into the Grid


TreeNewBeeing:
When I tried to implement a feature that I can drag an element from outside into the grid. 

I used onDrop event to assign the new layout variable to props of the grid. And when I tried to drag the new element from outside and place it on the top of the existing elements, it did not appears there but falls down to the bottom of them. When I am checking the data flow, I found that before I pass the props into the grid, the layout is correct on position with correct x, y. I am not sure what happens inside, causing this behavior. 

Details can be found through the codesandbox page.
[https://codesandbox.io/s/quirky-knuth-u78sp?file=/src/App.js](https://codesandbox.io/s/quirky-knuth-u78sp?file=/src/App.js)

When I drag from 'drag here', and place it on the top of yellow card, it falls down to the bottom of it.

vocko:
I will attach to this one as it seems to be a related issue, create a new one of it is not.

There seem to be multiple issues with dragging items from the outside. One is the placement described above, another is the fact you can't drop the item on the shadowed target area (these two might be actually related, but I'm not sure). Cursor appears to be `not-allowed` and you have to move your mouse out of the shadowed area to drop it, otherwise the `ondrop` handler is not called. This is not very nice user experience.

You can verify this behavior on the link above or also in this demo:
https://strml.github.io/react-grid-layout/examples/15-drag-from-outside.html

trace59:
> When I tried to implement a feature that I can drag an element from outside into the grid.
> 
> I used onDrop event to assign the new layout variable to props of the grid. And when I tried to drag the new element from outside and place it on the top of the existing elements, it did not appears there but falls down to the bottom of them. When I am checking the data flow, I found that before I pass the props into the grid, the layout is correct on position with correct x, y. I am not sure what happens inside, causing this behavior.
> 
> Details can be found through the codesandbox page. https://codesandbox.io/s/quirky-knuth-u78sp?file=/src/App.js
> 
> When I drag from 'drag here', and place it on the top of yellow card, it falls down to the bottom of it.

Hello, I also encountered this problem, is there a good solution?

trace59:
> I will attach to this one as it seems to be a related issue, create a new one of it is not.
> 
[...]

PR #1333, opened 2020-11-16
Firefox dragover

Charles-Lamoureux:
This is picking up from a separate PR that is stale: https://github.com/STRML/react-grid-layout/pull/1053
There has been a related issue opened since the original PR: https://github.com/STRML/react-grid-layout/issues/1324

If you try to drop an item over a placeholder in Firefox, the onDrop function is not invoked.

I found that with the current state of the project, it was only necessary to add preventDefault() to the event when using Firefox to resolve this.


STRML:
Thanks!

vocko:
@STRML Could an npm release be created with this fix, please? And thanks for the whole project, I really love your library!

STRML:
Yep - released as  `1.2.0`.

vocko:
> Yep - released as `1.2.0`.

Thanks heaps mate!

sayo96:
I'm experiencing something strange. Look at this issue - https://github.com/react-grid-layout/react-grid-layout/issues/1838

PR #1350, opened 2020-12-03
fix: when i dragend a item, onLayoutChange trigger twice.

yangguansen:
setState(layout) can trigger onLayoutMaybeChanged, so I think there should't manual call onLayoutMaybeChanged

example or test case.


STRML:
Sorry this isn't going to make it in as written. It is not clear what you are trying to fix and simply commenting out a line won't do.

Issue #1352, opened 2020-12-04
calcXY uses Math.round resulting in item popping from halfway in a column to the next column


Jarreddebeer:
In the example [Drag from Outside](https://strml.github.io/react-grid-layout/examples/15-drag-from-outside.html) the droppingItem pops to the adjacent column when the mouse reaches the middle of it. The effect is subtle in the example because the panel is small, but more jarring to the user when items are larger.

In my opinion, the more natural behaviour is to drop items to the column in which the mouse is hovering. It can be achieved in calcXY by replacing Math.round with Math.floor.

PR #1353, opened 2020-12-04
CalcXY uses Math.floor instead of Math.round so item doesn't pop to n…

Jarreddebeer:
[Related issue](https://github.com/STRML/react-grid-layout/issues/1352)

By replacing Math.round with Math.floor the droppingItem exhibits more natural behaviour when dragging into columns

### DIFF
--- a/lib/calculateUtils.js
+++ b/lib/calculateUtils.js
@@ -107,8 +107,8 @@ export function calcXY(
   // l - m = x(c + m)
   // (l - m) / (c + m) = x
   // x = (left - margin) / (coldWidth + margin)
-  let x = Math.round((left - margin[0]) / (colWidth + margin[0]));
-  let y = Math.round((top - margin[1]) / (rowHeight + margin[1]));
+  let x = Math.floor((left - margin[0]) / (colWidth + margin[0]));
+  let y = Math.floor((top - margin[1]) / (rowHeight + margin[1]));
 
   // Capping
   x = clamp(x, 0, cols - w);
--- a/test/spec/lifecycle-test.js
+++ b/test/spec/lifecycle-test.js
@@ -81,7 +81,7 @@ describe("Lifecycle tests", function () {
           h: 1,
           w: 1,
           x: 1,
-          y: 4,
+          y: 3,
           static: false,
           isDraggable: true
         });


STRML:
Given that `calcXY` is used all over the codebase, I'm not sure we want to change it here just to make this single usecase a little more natural.

Jarreddebeer:
Perhaps I misunderstand the extent of the method, it looks like it's used within 4 onDrag events. The test case results remained unaffected except for the dropping-item test case which this intends to affect.

'A little more natural' is an understatement in a sense. The behaviour of dropping-item drag is inconsistent with on-grid drag and unusable for respectably sized panel widths. I'd advocate the demo example should use larger width panels to portray the true behaviour because I did not notice it until it was implemented on a client project, and had to resort to another solution.

STRML:
@Jarreddebeer that's interesting. Can you point to a codesandbox where this issue is more obvious? I haven't seen it really affect day to day use in my use cases, but I'm happy to change it if there's a good reason.

Jarreddebeer:
Sure, I should perhaps have put up a visual initially sorry. The behaviour can be seen on the [drag from outside](https://react-grid-layout.github.io/react-grid-layout/examples/15-drag-from-outside.html) demo. Here i've zoomed in to mimic a 3 or 4 column grid and resized the panels to mimic w: 1, h: 1

output

The `Math.round()` in calcXY is mapping [0.5, 1.0] -> 1.0 which results in the item popping to the adjacent panel. Whereas regular dragging does not experience this. Using `Math.floor` instead seems to achieve the desired behaviour (when I tested)

LoganDark:
> Given that `calcXY` is used all over the codebase, I'm not sure we want to change it here just to make this single usecase a little more natural.

This is absolutely what you want and not `Math.round` - floor ensures that all values from 0..1 are in the first column, while rounding means only values from 0..0.5 are in the first column, and values from 0.5..1.5 are in the second (weird).

STRML:
Thanks. Will review this again.

STRML:
The real issue here is how droppability works - it is significantly more intuitive when you are dragging from the left side of the object. This is not exhibited with general drag/drop within the grid.

Unfortunately, this solution negatively affects general drag/drop, causing it to feel significantly less intuitive and more shifted to the right. For this reason I can't accept the PR as written, and would instead point you toward the logic in `[ReactGridLayout.onDragOver()](https://github.com/react-grid-layout/react-grid-layout/blob/1547a474a5778ee1ae8cd4479b19615f0d46c041/lib/ReactGridLayout.jsx#L622)`  - it should be calculating the center or left edge of the droppable item and calculating off that, rather than the mouse position.

STRML:
Unfortunately [there is no way to actually get the element being dragged](https://stackoverflow.com/a/23416174) on a `dragover` event which makes this significantly more difficult.

timadevelop:
Maybe it's against some standards, but I've found a way to make it work properly.

RGL calls `onDropDragOver` callback with react `SyntheticEvent` instance and uses the same event instance for calculating XY of a dragged item.
Modifying `SyntheticEvent.nativeEvent` itself is not allowed by web standards, but `SyntheticEvent` provided by react is a regular mutable JS object. It can be "patched" with desired values.

The code below will work as long as RGL reads `layerX/layerY` data after calling `onDropDragOver` here  https://github.com/react-grid-layout/react-grid-layout/blob/master/lib/ReactGridLayout.jsx#L652-L663

@STRML thank you for this lib. Hopefully `nativeEvent` reading will stay the same way in the future, after the callback invocation.

> Unfortunately there is no way to actually get the element being dragged on a dragover event which makes this significantly more difficult.

This can be handled out of the RGL. In my use case, I don't even need that element, I store the metadata associated with it in my component state/context.

```js

  const onDropDragOver = useCallback(se => {
    if (!lastDraggedToolboxItem) return false;

    const { type, size } = lastDraggedToolboxItem;
    if (!size) return false;

    // This nativeEvent patch fixes the following issue:
    // https://github.com/react-grid-layout/react-grid-layout/pull/1353#issuecomment-873202846
    //
    // Simply offset dragging position with 1/2 of "future" item width.
    // Do not use bounding box / global size of dragged DOM element here, offset using the size
    // of "future" GridItem that will be inserted *within* RGL view itself.
    se.nativeEvent = {
      ...se.nativeEvent,
      layerX: se.nativeEvent.layerX - (size.w * cellDimensions.wPx / 2),
      layerY: se.nativeEvent.layerY - (size.h * cellDimensions.hPx / 2),
    };

    return { w: size.w, h: size.h };
  }, [lastDraggedToolboxItem])

```

Jarreddebeer:
> The real issue here is how droppability works - it is significantly more intuitive when you are dragging from the left side of the object. This is not exhibited with general drag/drop within the grid.
> 
> Drag from left is fairly intuitive:
> 
>  drag-from-left.mov 
> Drag from right is not:
> 
>  drag-from-right.mov 
> Unfortunately, the solution in this PR negatively affects general drag/drop, causing it to feel significantly less intuitive and more shifted to the right. For this reason I can't accept the PR as written, and would instead point you toward the logic in [`ReactGridLayout.onDragOver()`](https://github.com/react-grid-layout/react-grid-layout/blob/1547a474a5778ee1ae8cd4479b19615f0d46c041/lib/ReactGridLayout.jsx#L622) - it should be calculating the center or left edge of the droppable item and calculating off that, rather than the mouse position.

This analysis is slightly misleading and not inline with that of my original PR, for the following reason: 

In your example you have cards which span two columns. This creates the illusion that the behaviour is natural, but it's fake, your grid-x resolution is twice that of the cards. Compare it to my example where the cards are take up a single column (not two as in your case).

It is not natural to have a grid with X=6 and cards with W=2, which allows for cards to sit over 'half a column'.
Instead, people would use a grid with X=3 and cards with W=1.

Math.floor is a less error prone mapping for the range of columns. It does not even need Math.min to ensure the card does not round up beyond the number of columns

Jarreddebeer:
> Unfortunately, the solution in this PR negatively affects general drag/drop, causing it to feel significantly less intuitive and more shifted to the right.

I'm not sure where this is seen, the solution in the PR literally shifts the cards more to the left.

PR #1367, opened 2021-01-01
Fix position bug (drag in scaled mode)

Valkyr12:
We have position bug when dragging GridItem in scaled mode. 

[CodeSandbox example](https://codesandbox.io/s/eager-leaf-kmedd)

This is old bug, and its have been already fixed. But its looks like it have been fixed twice, first in the react-grid-layout and second in react-draggable (may be in react-scalable too). As a result, we again have a positioning error.


STRML:
Fixed in 1157e13639f4d83ede6c8524c1b0ef1c80e5d6cb

Issue #1337, opened 2020-11-21
ResizeHandle not working


Abhijit-H-Kulkarni:
Upgraded to latest version but still the resizeHandle doesn't seem to work. It doesn't replace the resizeHandle.

CodeBox : https://codesandbox.io/s/resizable-111-forked-1s770

JerryLeeCS:
I am having the same problem.

valtism:
Yeah, getting this too. It seems to work with a custom component if it is give then class name of `react-resizable-handle`, but then the default styles get applied as well.

valtism:
Ok, figured this out. If you specify a classname using the `draggableHandle` prop, then ensure that your custom element has that same classname, it will work.

Issue #1317, opened 2020-10-20
resizeHandle prop does nothing


kylesureline:
The `resizeHandle` prop (as mentioned [here](https://github.com/STRML/react-grid-layout#grid-layout-props)) doesn't appear to exist on the ReactGridLayout component. It does seem to be implemented on the GridItem component and ReactGridLayoutPropTypes, however.

Also, I noticed that this example is producing a 404: https://strml.github.io/react-grid-layout/examples/17-resizable-handles.html

typeetfunc:
version with this changes not released yet https://github.com/STRML/react-grid-layout/pull/1303

kylesureline:
🤦 my bad. Well, I'll patiently wait for the release. Thanks!!

dwb1994:
shouldn't appear on readme.md if the changes not released 😡

davidbejarcaceres:
Any news?  resizeHandle is doing nothing
resizeHandle: resizeHandle={axis => {
      return <TdIcon axis={axis} medium iconName={'arrowDown'} />;
    }}

curtislin7:
+1, a new release with this feature would be much appreciated. Thanks guys!

PR #1368, opened 2021-01-02
Add instructions on resizeHandle prop usage.

valtism:
Usage of `resizeHandle` prop was unclear. Seen in #1337 and #1317


valtism:
Incorrect solution

PR #1371, opened 2021-01-04
fix: flow type check error (DragOverEvent)

barrychen38:
Correct `classList` type from console

image


STRML:
Thanks. The type here really isn't exact. I need to spend more time with it. However, the type error is fixed in 53b31606fa42b3a76ebd796d3bb0f165f5aa6d9b

Issue #1376, opened 2021-01-12
Items not compacting properly in responsive mode, weird glitch when dragging items


drixie:
See codesandbox - https://codesandbox.io/s/hardcore-dew-0bzip?file=/src/App.js

First issue - When the breakpoint is set to md or lg, Box "a" does not fill up the space adjacent to "e" and below "b" as expected. What can I do to make this work? It would also be nice to have to boxes display in accordance to how they are structure (e.g. "a" should be displayed first, followed by b, and c... in a left to right order). 

Second issue - Whenever, i try to drag items, and the item being dragged overlaps existing items, there is shaking effect during which the structure of all the items is temporarily disrupted until the drop has been made. How do I fix this issue? It is not nice from a UX perspective.

drixie:
Issue #1 seems to be a long running issue - https://github.com/STRML/react-grid-layout/issues/66

The only solution I have found so far is to declare the initial layout and account for all the [x, y, w, h] values. That said, I am not sure how well it will hold when users drag things around in ways you never thought possible. 

Issue #2 happens only in horizontal compaction. Does not occur when vertical compaction is set

drixie:
Detailed discussion on issue #1 from 2018 - https://github.com/STRML/react-grid-layout/issues/784

Any takers on solving this in 2021? 😄

s4m3:
@drixie thanks for digging up the latest discussions on that. My main concern is issue 2 right now. Will have a look into the code today, to see what is causing this.

s4m3:
@drixie just an FYI: I did work on it for a couple of hours and was able to isolate one of the issues, but then more pressing issues came up, so this is put on hold. Will get back to it over the week however.

s4m3:
@drixie I opened a Pull Request that solves the second issue, but I am not sure what the response time is here with the owner. Meaning how long it takes until he looks at these Pull Requests. I see a lot of stale Pull Requests as well. I guess we can only be patient...

drixie:
@s4m3 Thanks a lot for the prompt action on this. Let's hope the owner reviews your PR soon. If we don't get a response in two weeks, do you mind setting up an alternate repo / npm package so I can test?

s4m3:
@drixie you can checkout this one locally and try it standalone:
https://github.com/s4m3/react-grid-layout/tree/bugfix/horizontal-compact-mode-issues

Issue #1372, opened 2021-01-05
Horizontal Compaction feels buggy


Abhijit-H-Kulkarni:
RGL with vertical compaction feels more rigid when compared to horizontal compaction. When the compaction is horizontal moving elements in the grid results in some weird positioning of elements which is not expected. Also moving the grid items feels buggy compared to vertical compaction. For my use-case I needed both horizontal and vertical compaction, tried to do it with compactType='horizontal' and verticalCompact=true, but seems that the horizontal compaction is causing issues. Basically if there is some space available for a grid to occupy that space is not being utilized rather it will be pushed to the end. Is anyone having a similar use-case as mine and facing this issue?

Codebox Example : https://codesandbox.io/s/currying-water-xvdtn?file=/src/ShowcaseLayout.js

monmush:
I am facing the same problem

justin-restoguru:
I'm facing the same problem, can someone help me please

PR #1390, opened 2021-01-25
Horizontal Compact Mode issues

### LINKED ISSUE #784: Several bugs with horizontal compactType

s4m3:
Fixes #784 #1372 #1376

Horizontal compact mode should not compact vertically by reducing l.y by using the bottom function: It reduces the container and triggers items of the layout with the highest y values to reposition themselves because the repositioning is mixed with the horizontal compaction. So in my opinion, this one line of code needs to be removed (as done in this Pull Request).

Before:

https://user-images.githubusercontent.com/2151167/105766228-c5c9d900-5f59-11eb-8ee0-f9d34e48d88f.mp4

After:

https://user-images.githubusercontent.com/2151167/105766241-cbbfba00-5f59-11eb-81e2-7f9c212659f0.mp4

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -79,14 +79,23 @@ const DEBUG = false;
  * Return the bottom coordinate of the layout.
  *
  * @param  {Array} layout Layout array.
+ * @param  {CompactType} compactType to identify the orientation for the
  * @return {Number}       Bottom coordinate.
  */
-export function bottom(layout: Layout): number {
+export function bottom(layout: Layout, compactType: CompactType): number {
+  const compactH = compactType === "horizontal";
   let max = 0,
-    bottomY;
+    bottomPos;
+
+  // we need some default, so take vertical?
+  let getBottomPos = item => item.y + item.h;
+  if (compactH) {
+    getBottomPos = item => item.x + item.w;
+  }
+
   for (let i = 0, len = layout.length; i < len; i++) {
-    bottomY = layout[i].y + layout[i].h;
-    if (bottomY > max) max = bottomY;
+    bottomPos = getBottomPos(layout[i]);
+    if (bottomPos > max) max = bottomPos;
   }
   return max;
 }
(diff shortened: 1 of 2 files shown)


s4m3:
I would argue the test that fails above is not what should be expected, why would y be reduced if compact mode is horizontal?

s4m3:
image
 Is there some special node version used or what is the issue here?

s4m3:
@STRML any help would be appreciated (regarding running the tests as well as reviewing)

s4m3:
@STRML figured out the test issues: i used an old node version. So tests are running and while reading through the changes I found my initial error reading the code. 
I updated the Pull Request.
I think the real issue that caused the horizontal compact mode problems is the trial to reduce l.y in case of horizontal compact mode: This reduces the container and triggers items in the more bottom part of the layout to reposition themselves. So in my opinion, this one line of code needs to be removed (as done in this Pull Request).

s4m3:
ping @STRML 
Any chance you can take a look at this one? Or who are the other people who can review and accept PRs?

STRML:
Thanks, this looks decent, could you please update the videos though?

s4m3:
@STRML thanks for looking into it. I am not sure what you mean by updating the videos. What do you need?

PR #1393, opened 2021-01-28
Fixing the buggy drag container movement in scaled mode.

JitendraCh-Kore:
When parent is scaled (like transform: scale(0.5)) then even after applying transformScale props, it was not proper dragging experience. After removing the above line it's working proper.

example or test case.


jamesonhill:
I am also having this issue. 

https://user-images.githubusercontent.com/25965865/108518834-07b71800-7297-11eb-82ee-20c7980ba0a9.mov

ronyland:
I fixed it by updating the react-draggable from version 2.2.2 to version 4.4.3

Issue #1394, opened 2021-01-29
Dynamic `droppingItem` prop or expose `onDragOver`


JamesRamm:
Hi

I am using RGL with drag and drop to drag items from a sidebar onto the grid. Each item has it's own minW/minH which is carried in the drag data (`dataTransfer`). This works great but I would like to make the `droppingItem` overlay also respect the minimum size. 

The only way I can see to do this would be possibly to expose the onDragOver event, or rather, accept a function prop which will be called at the start of the `onDragOver` handler in ReactGridLayout, so that we may dynamically update `droppingItem`.
Or `droppingItem` could be that function - e.g. at the start of `onDragOver` if `droppingItem` is a function, call it with the event and expect to receive the `droppingItem` object as the return value.

I am happy to propose a patch, but was wondering if some kind of control like this is already possible?

JamesRamm:
No response on this? I don't consider it stale, just that no one seems to be maintaining this project...

sonmansu:
Did you find any solution? same problem with me.

B33fb0n3:
Dynamic configuration is possible, but you **can't access** the data, that you entered while setting the data:
```tsx
e.dataTransfer.setData("text/plain", "data"); // or stringified data string
```
Am I just blind or is there no option to get this data inside the ``onDropDragOver``-Event to determine the size of the element?

PR #1395, opened 2021-01-29
#1394 - Dynamic configuration of droppingItem

JamesRamm:
For issue #1394 
Idea here is to allow the `droppingItem` to be dynamically adjusted based on what is being dragged over.
Question - maybe we should prevent the `i` property from being overwritten (and limit it to `w` and `h`). I am not sure but dynamically changing `i` may cause issues for removing the placeholder. I don't see any use case for needing to change `i` anyway.

Example, in my particular use case the `onDropOver` handler may look like:

```js
const getDragData = e => e.dataTransfer.getData('dragData');
const evolveKeys => ({ minW: w, minH: h }) => ({ w, h })
const onDropOver = R.tryCatch(R.compose(evolveKeys, JSON.parse, getDragData),  () => ({ }));
```

### DIFF
--- a/lib/ReactGridLayout.jsx
+++ b/lib/ReactGridLayout.jsx
@@ -602,7 +602,11 @@ export default class ReactGridLayout extends React.Component<Props, State> {
       width,
       containerPadding
     } = this.props;
-    const finalDroppingItem = onDragOver ? { ...droppingItem, ...onDragOver(e) } : droppingItem;
+    // Allow user to customize the dropping item or short-circuit the drop based on the results
+    // of the `onDragOver(e: Event)` callback.
+    const onDragOverResult = onDragOver?.(e);
+    if (onDragOverResult === false) return;
+    const finalDroppingItem = { ...droppingItem, ...onDragOverResult };
     
     const { layout } = this.state;
     // This is relative to the DOM element that this event fired for.
--- a/lib/ReactGridLayout.jsx
+++ b/lib/ReactGridLayout.jsx
@@ -594,6 +594,7 @@ export default class ReactGridLayout extends React.Component<Props, State> {
 
     const {
       droppingItem,
+      onDragOver,
       margin,
       cols,
       rowHeight,
@@ -601,6 +602,8 @@ export default class ReactGridLayout extends React.Component<Props, State> {
       width,
       containerPadding
     } = this.props;
+    const finalDroppingItem = onDragOver ? { ...droppingItem, ...onDragOver(e) } : droppingItem;
+    
     const { layout } = this.state;
     // This is relative to the DOM element that this event fired for.
     const { layerX, layerY } = e.nativeEvent;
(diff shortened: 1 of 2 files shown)


JamesRamm:
Hello maintainers? Any reply on this PR. It is not stale - still perfectly valid and compatible.

STRML:
Thanks, happy to merge this if you could add a brief test & README update.

STRML:
@JamesRamm something to think about - we usually allow returning `false` from handlers like this, too, so that the user can choose to disable the functionality based on some logic. I'll add a suggestion, let me know if you think it is useful.

lib/ReactGridLayout.jsx
```diff
@@ -594,13 +594,16 @@ export default class ReactGridLayout extends React.Component<Props, State> {
...
       droppingItem,
+      onDragOver,
       margin,
       cols,
       rowHeight,
       maxRows,
       width,
       containerPadding
     } = this.props;
+    const finalDroppingItem = onDragOver ? { ...droppingItem, ...onDragOver(e) } : droppingItem;
```
STRML, line 605:
```suggestion
    // Allow user to customize the dropping item or short-circuit the drop based on the results
    // of the `onDragOver(e: Event)` callback.
    const onDragOverResult = onDragOver?.(e);
    if (onDragOverResult === false) return;
    const finalDroppingItem = { ...droppingItem, ...onDragOverResult };
```
```

**Comment by wuzhong1030** [comment]
Thank you!
It's commit very helpful for me.

**Comment by STRML** [comment]
Apologies, updated in pr/1395 and accidentally not pushed before merge. 415172f contains these changes.

**Comment by sayo96** [comment]
@JamesRamm  I'm unable to get the dragData through event.dataTransfer inside the onDropDragOver . Is there any way to implement this

**Comment by ekuzu** [comment]
@JamesRamm is it possible to get dataTransfer object inside onDropDragOver event?

**Comment by JamesRamm** [comment]
@sayo96  @ekuzu 
The event is `onDragOver`, not `onDropDragOver`

**Comment by sayo96** [comment]
@JamesRamm I need the event `onDropDragOver`. If possible please look at this issue i ran into in firefox https://github.com/react-grid-layout/react-grid-layout/issues/1841

**Comment by sonmansu** [comment]
> @sayo96 @ekuzu The event is `onDragOver`, not `onDropDragOver`

`GridLayout` does not have `onDragOver` props.
How can I get `dataTransfer` object inside `onDropDragOver`or `onDragOver`?

**Comment by B33fb0n3** [comment]
If you add data inside the droppable-element, that data is not accessable inside the onDragOver (onDropDragOver prop).

Add Data:
```jsx
onDragStart={e => {
 e.dataTransfer.setData("text/plain", JSON.stringify(component.blockData));
}}
```

But if you try to access this data via 

```jsx
const eventData = e.dataTransfer.getData("text/plain");
```
It returns null ^^

If you access it the same way while handling the drop event, you get the data

szabikr:
Any updates on this?

I would also like to get the `dataTransfer` data from the event in the `onDropDragOver` handler but get an empty string.

Thank you!

mm105:
Am I missing something or there is still no option to have size of dragged item while dragging over grid to dynamically determine size of palceholder?

PR #1411, opened 2021-03-02
fix(callbacks): ensure onDragStop always fires if onDragStart did

### LINKED ISSUE #1341: onDragStart fires before any mouse movement
I have clickable elements in my Layout children that I render inside of the GridLayout component. To avoid launching new tabs or other events I am disabling click events while the children are dragging. I am hooking into the onDragStart and onDragStop event props to set my isWidgetMoving context object. The bug is that the onDragStart fires on mouse click instead of mouse down and some pixel movement.

**Live Demo (Stackblitz)**
https://stackblitz.com/edit/react-grid-layout-bug-dragstart?file=index.tsx

Recreation Steps:
1. Click the button without moving the mouse.
2. Notice the button text updates to "moving" even though there is no dragging.

### LINKED ISSUE #1401: onDragStart fired when draggable container is merely clicked, but not onDragStop
Example: https://codesandbox.io/s/angry-pond-dj3tl?file=/src/ShowcaseLayout.js:1829-1937

If you just click (not drag) on the draggable container you see the log from `onDragStart` in the console, but not from `onDragStop`.

Expected: Either none of the events, or both should be fired, so you can programmatically catch both beginning and end of drag drop.

This bug was introduced with this PR: https://github.com/STRML/react-grid-layout/pull/1023

Workaround: Use `onDrag` instead of `onDragStart`

STRML:
Fixes #1341
Fixes #1401

This problem comes about because we short-circuit some logic in `ReactGridLayout#onDragStop` when no active drag was present, but we really should fire the user-facing callback, otherwise they *must* listen to `onDrag` instead and manage their own state if they want to use it effectively.

Alternatively we could wait until the first actual drag to call the callback (rather than mousedown). I'm interested in hearing pros/cons to this if anyone has any.

cc @djmisterjon @breautek @piltorious @DNosralla


breautek:
> Alternatively we could wait until the first actual drag to call the callback (rather than mousedown). I'm interested in hearing pros/cons to this if anyone has any.

In my use this, I think this would be better and I don't think this quite solves the problem with #1341 
I'll provide an example. In my use case we have a widget that has two actions, a click action and the drag action.

The problem I face (and the same as the OP of https://github.com/react-grid-layout/react-grid-layout/issues/1341 I believe) is that when you do a click, the drag initiates on mouse down, creating a move effect like shown in the attached gif. We would like to only initiate a drag once the mouse starts moving so the widget only is "grabbed" when the user moves the mouse, not when they do a click. This is the premise of the threshold proposal I laid out in https://github.com/react-grid-layout/react-grid-layout/issues/1341#issuecomment-789014024

This way, on click the widget won't move until the mouse actually starts moving.

output

If there is anything I could do to help, do let me know.

djmisterjon:
Both solution suite me.
But i do not have enough experimental cases to tell what will be the better way.
Maybe provide a hight setting to offer the two bheiavior fix.

To appeal to everyone, sometime it a good idea to provide a small Config API with flag at high level to allowing different behavioral.
ex:
```js
//user
import GridLayout,{RGLConfig} from 'react-grid-layout';
RGLConfig._FORCE_DRAG_STOP_ =true;

//core RGL
 if (RGLConfig._NO_DRAG_POINTER_DOWN_ ) {
   // ... stuff
 }
if (!this.state.activeDrag && RGLConfig._FORCE_DRAG_STOP_ ) {
   // ... stuff
 }
```
example libs gsap do this for allow people to manage breakchange and also cover all scenario usage.
https://greensock.com/docs/v3/GSAP/gsap.config()



> Alternatively we could wait until the first actual drag to call the callback (rather than mousedown). I'm interested in hearing pros/cons to this if anyone has any.

yep i also remember facing the issue @breautek showed upper.

It can happen from first calculations when we customize the height, compact and use `useCSSTransforms={false}` behavior.
_mouse click can be see left with yellow circle and right by red and greed log_
a3faga3ga3fg

the issue are not present with `useCSSTransforms={false}`  because calculation seem ok the first rendering.
But the fast click also fires `onDragStart` without `onDragStop`.
a3faga3ga3fg-15af

conclusion
* force fires `onDragStop` when `onDragStart` emitted is a good solution for me. 
* emit only `onDragStart` when first drag is detecting can be also a good solution for me , but can be problematic if the mouse down make small move of the elements.

both could be offered with a high level API settings. `GridLayout.RGLConfig`
or if no API setting can be made,  i will have a preference for the first solution, because I already know behavior.
This is my personal opinion for my personal use.
thanks for all @+

djmisterjon:
debug layers in 3d will also reveal more the issu.

piltorious:
To bring this back to the original issue. onDragStart should not fire until the user is actually moving the widget. A click should not start that process alone. A click and some minimum pixel movement x or y should. The stackblitz that I put on the original post highlighted it well what happens if you move the widget vs just clicking.

STRML:
Fair enough @piltorious, I agree with you.

STRML:
Updated @piltorious, please see the new description & changelog as well.

piltorious:
@STRML Are there other contributors to this library that will review the PR or am I it?

STRML:
Pinging @JamesRamm who has been active lately

stefanbugge:
Looking forward to getting this fix out. 
Any chance of getting someone on the team to review this soon? 
Thank you.

djmisterjon (approved):

sigalor:
Ping

STRML:
Could use additional review on this.

SaeedZhiany:
Hi everybody,

Is there any update on this? will this PR get merged?

Thanks

SaeedZhiany:
Hi,

Is there any update on this? will this PR get merged?

Thanks

DavidHenri008:
Any development concerning this PR?

SaeedZhiany:
a

manuelnogueiramov:
Can we have an estimation on when can this PR be delivered? `ignoreClickOnly` seems like would be very helpful :)

jnavb:
is there any update to this PR? 

Thanks for this amazing package!

lexnim:
waiting for this merge

bruce-c-liu:
Ping.

sidrak19:
ping :)

Konders:
Ping

chenghaoc:
ping

djmisterjon:
ping or bing ? https://youtu.be/m5uizpXJInY?t=7

sanderdsz:
ping it!

STRML:
Unfortunately this needs a pretty big rebase and I don't have time for it. If anyone else does, please ping me and I'll pull the changes in.

yf-yang:
@STRML I just rebased the code, but 2 tests still fail. I think `simulateMovementFromTo` function isn't working properly so `GridItem.moveDroppingItem` returns earlier than expected, but I don't know what is the correct way to mock it. Would you like to take a look?

STRML:
@yf-yang I'd be happy to look if you post exactly what you're dealing with.

nafishey:
2024年了这个问题还没有解决吗

Issue #1442, opened 2021-04-07
Drag from Outside issue in Chromium browsers


TomasKomprs:
Hello,
I have seen similar issues here, but all are closed and not fixed.

Issue is "only" presented in Chrome / Chromium browsers. If you drop item from outside, it sometimes didn't drop.
I have included fix for Firefox and D&D/drop from outside works 100% correct. "Only" in chromium browser is this issue. (Chrome and Brave for example)

We can see this issue in your demo too. So i don't have to reproduce it in codesandbox. Just try it a few times. Drop it in different places, different speed etc.
https://react-grid-layout.github.io/react-grid-layout/examples/15-drag-from-outside.html

I tried figure out what it could be, but so far I don't have any idea. Don't have much time look at this. Can someone check this out, please? I think this is a crucial issue.

I'm running React 17+, react-grid-layout 1.2.4

STRML:
Thanks, fixed in `1.2.6`.

PR #1448, opened 2021-04-15
fix: add preventDefault() for chrome dropping from outside sometimes …

kiddos:
I am having this issue: #1442. 
Peek 2021-04-16 00-31
Peek 2021-04-16 00-32
(the above was modify from `15-drag-from-outside.jsx`)

This sometimes happens. My friend is on Mac, who doesn't seem to have this problem.
I am on Linux Chrome 89.0.4389.114.
Adding `e.preventDefault()` in onDragOver(already has that) and onDragEnter event works for me.

### DIFF
--- a/lib/ReactGridLayout.jsx
+++ b/lib/ReactGridLayout.jsx
@@ -701,7 +701,9 @@ export default class ReactGridLayout extends React.Component<Props, State> {
     }
   };
 
-  onDragEnter: () => void = () => {
+  onDragEnter: (e: Event) => void = (e: Event) => {
+    e.preventDefault();
+
     this.dragEnterCounter++;
   };
 


STRML:
Fixed in 0ab31d4

kiddos:
Cool thanks.

shivampip:
Hy @kiddos , how did you set **w** and **h** of item being dragged.
By default, both are set to 1.

kiddos:
I did the following:
1. add `droppingItem` for `<ResponsiveReactGridLayout>`
2. `droppingItem` would look like `{i: 'new id', w: 6, h: 6}`
3. add `onDrop` handler for `<ResponsiveReactGridLayout>`
4. `onDrop` would add the `droppingItem` to the layout
5. adding the `droppingItem` into the layout is kinda tricky. The simplest strategy is to increment all the items with `y >= current droppingItem.y` y current droppingItem's height, and let itself realign to top. However, you can also deploy other strategies on your own.
6. One final note is that the next  `onLayoutChange` should be triggered after the `onDrop`, because if it was triggered before `onDrop`, the `onLayoutChange` does not see the `droppingItem` yet and the `droppingItem` will not be added at the end. When I tried using Redux, `onLayoutChange` was sometimes triggered before `onDrop`, and the `droppingItem` does not get added. So the simplest way to get rid of this is to add a flag `this.dropping = true` in `onDrop`. add a skip in `onLayoutChange` using that flag, and `this.dropping = false` in render.

My example here: [https://stackblitz.com/edit/react-grid-layout-drag-in?file=src/App.js](example)
Note: the version of `react-grid-layout` is before this fix. So sometimes the grid item doesn't get added on Chrome.

PR #1470, opened 2021-05-13
Overlap grid item

gokulb9993:
We had a use case where we wanted to place one grid over the other. On top of this we have added feature like "Bring forward, Bring backward, Bring one step forward, Bring one step backward" the grid items.

**Before:**

https://user-images.githubusercontent.com/42829533/118166085-af7d5580-b442-11eb-8255-2a9aed045d20.mov

**After:**

https://user-images.githubusercontent.com/42829533/118166136-bc9a4480-b442-11eb-8e61-3fd0603a6ea8.mov


STRML:
Pretty simple patch, could you add some tests?

I am thinking about refactoring both the compaction and the collision algos into classes to make them pluggable. In that situation, you could replace the collision implementation with a noop and you'd basically be done. How does that sound for an idea?

gokulb9993:
@STRML Added tests

STRML:
Thanks, I meant an actual test though (that folder is misnamed), in the test/spec folder.

gokulb9993:
@STRML Added functional test case. Please let me know if anything is missed

STRML:
Thanks!

klifa90:
Any news on when this is being released?

Issue #1485, opened 2021-06-02
Nested grids drag drop doesn't dispose when hovering over nested grid


rogerfar:
I couldn't reproduce this on codesandbox because dragging doesn't seem to work there, so I made a case on stackblitz:
https://stackblitz.com/edit/react-qeuac3?file=src%2FApp.js

When you drag the element into the first grid on the right side and the drag it to the left between 0 and 1 you can see that the placeholder doesn't disappear.

It will also trigger 2 drop events, which can be fixed by adding `event.stopPropagation();`, but this causes the ghost overlay to remain and only to disappear when starting another drag.

rogerfar:
I added 
`e.preventDefault();
e.stopPropagation();`

To onDragLeave, onDragEnter and onDrop and that seems to fix it.

@STRML do you see any issues with that, I can create a small PR.

thomaspessato:
hey @rogerfar, do you happen to have that fixed example?

rogerfar:
I can have a go at trying to get the patch in, it's not that easy as the current NPM package is a bit out of date from master.

rogerfar:
@thomaspessato I updated my example here: https://stackblitz.com/edit/react-qeuac3?file=src%2Freact-grid-layout.js.

The only thing I did here was take the current master, build it and pasted the minified output in `react-grid-layout.js`. Then I added `e.stopPropogation`:

```
F(X(t), 'onDragLeave', function(e) {
                  e.preventDefault(),
                    e.stopPropagation(),
                    t.dragEnterCounter--,
                    0 === t.dragEnterCounter && t.removeDroppingPlaceholder();
                }),
                F(X(t), 'onDragEnter', function(e) {
                  e.preventDefault(), e.stopPropagation(), t.dragEnterCounter++;
                }),
                F(X(t), 'onDrop', function(e) {
                  e.preventDefault();
                  e.stopPropagation();
                  var r = t.props.droppingItem,
                    n = t.state.layout,
                    o = n.find(function(t) {
                      return t.i === r.i;
                    });
                  (t.dragEnterCounter = 0),
                    t.removeDroppingPlaceholder(),
                    t.props.onDrop(n, o, e);
                }),
```

rogerfar:
Can this be re-opened? This bug is pretty annoying and I added a PR for it.

willh33:
I need this as well

willh33:
@rogerfar Does drag and drop work with nested grids?

[...]

PR #1494, opened 2021-06-11
Add e.stopPropagation to onDragLeave, onDragEnter and onDrop to preve…

rogerfar:
Ref issue #1485

Fixes an issue when using a nested grid dropping an element from outside will trigger on both the child and parent element. Also the ref counter will be out of sync because it will use the parent too.


willh33:
Any news on when this will get merged? I'm trying to use nested grids and having some issues that I think this will resovle

KarlGong:
We really need this fix. Does anyone can merge it? @STRML

STRML:
Thanks. Will add a patch release for this.

PR #1512, opened 2021-07-06
Remove existing placeholder when onDropDragOver return false

ddequidt:
Resolves issue [1511](https://github.com/react-grid-layout/react-grid-layout/issues/1511)


STRML:
Thanks, solid PR.

PR #1517, opened 2021-07-23
feat: copy the PR where ignoreClickOnly prop is added 

oquirozm:
Copy from the original PR where ignoreClickOnly prop is added to prevent onDragStart from being incorrectly called on click and the change where the address the issue about onDragStop not being called everytime onDragStart is called.


oquirozm:
Sorry for the inconvenience, feel free to delete the PR, created it on the original repo instead of our fork.

Issue #1522, opened 2021-07-29
When I use droppingItem and transformScale:0.5 the widget not follow with mouse.


pingguotailang:
When I use droppingItem and transformScale:0.5 to drag over some element from outside. the widget not follow with mouse.
![bug gif](https://user-images.githubusercontent.com/41989206/127425446-2bf8553c-2a42-47db-bc23-41df7b6989de.gif)

PR #1530, opened 2021-08-04
Fix drop dragover not obeying transformScale

daveagill:
Fixes issue #1522 

The position given to us from the drag-drop event was not being scaled by the appropriate `transformScale`.


STRML:
Thanks!

daveagill:
No problem!  Any chance of publishing this as a new patch release? It would be a help if it were on npm.

pingguotailang:
Thanks!  Any chance of publishing this as a new patch release? It would be a help if it were on npm.

PR #1531, opened 2021-08-06
Fix dragover positioning when using CSS translation

daveagill:
__The Problems:__
When the grid is translated with a CSS transform the calculated dragover positions were being miscalculated within the grid.

This happens because `layerX` and `layerY` in the native event are not relative to the translated element, they are relative to the "layer" but the layer is not translated. So for example: If the grid is translated to (100px, 100px) and you then drag a new droppable into the grid at its top-left corner then `layerX` and `layerY` report (100,100) instead of (0,0) so both the placeholder and grid item are misplaced.

Also `layerX` and `layerY` are [not standard event fields](https://developer.mozilla.org/en-US/docs/Web/API/UIEvent/layerX) so differnet browsers may behave differently around this issue.

__The Solution:__
This PR implements a solution by measuring the final client (AKA on-screen) position of the grid (which takes into account the CSS transform) and compares that with `clientX` and `clientY` fields from the event to calculate the relative offset within the grid.

__Possible areas for improvement:__
1. This now calls `getBoundingClientRect` on every DragOverEvent. Unless we expect the grid to be moving at the same time then we could call it once and cache the result. Having said that I have not observed a perf impact.
2. This is quite similar to the implementation used in [`DragItem.onDragStart`](https://github.com/react-grid-layout/react-grid-layout/blob/9bf5942c06693ce31d88d8cf9849b1f56bb6c9b6/lib/GridItem.jsx#L432) but with some notable differences. I wonder if/how they could be common'd out and whether this PR should be factoring in `scrollLeft` and `scrollTop` too?
3. I noticed a [Firefox specific hack to fix event flickering](https://github.com/react-grid-layout/react-grid-layout/blob/9bf5942c06693ce31d88d8cf9849b1f56bb6c9b6/lib/ReactGridLayout.jsx#L604) which I suspect may also be resolved by this PR now that we use the better-defined `clientX` and `clientY` in screenspace and is no longer sensitive to whether the event target is the grid or a child of the grid.

### DIFF
--- a/lib/ReactGridLayout.jsx
+++ b/lib/ReactGridLayout.jsx
@@ -138,6 +138,8 @@ export default class ReactGridLayout extends React.Component<Props, State> {
 
   dragEnterCounter: number = 0;
 
+  ref = React.createRef();
+
   componentDidMount() {
     this.setState({ mounted: true });
     // Possibly call back with layout on mount. This should be done after correcting the layout width
@@ -657,9 +659,13 @@ export default class ReactGridLayout extends React.Component<Props, State> {
     const finalDroppingItem = { ...droppingItem, ...onDragOverResult };
 
     const { layout } = this.state;
-    // This is relative to the DOM element that this event fired for.
-    const { layerX, layerY } = e.nativeEvent;
-    const droppingPosition = { left: layerX / transformScale, top: layerY / transformScale, e };
+
+    const { clientX, clientY } = e.nativeEvent;
+    const dropTarget = this.ref.current?.getBoundingClientRect() ?? {left:0, top:0};
+    const offsetX = (clientX - dropTarget.left) / transformScale;
+    const offsetY = (clientY - dropTarget.top) / transformScale;
+
+    const droppingPosition = { left: offsetX, top: offsetY, e };
 
     if (!this.state.droppingDOMNode) {
       const positionParams: PositionParams = {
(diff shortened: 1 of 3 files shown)


lib/ReactGridLayout.jsx
```diff
@@ -755,6 +761,10 @@ export default class ReactGridLayout extends React.Component<Props, State> {
 
   render(): React.Element<"div"> {
     const { className, style, isDroppable, innerRef } = this.props;
+    
+    if (innerRef) {
+      innerRef.current = this.ref.current;
+    }
```
daveagill, line 767: Hmm, maybe this should be moved to `componentDidMount` to ensure the `innerRef` hook is sync'd after render 🧐

kiddos:
yeah, I really need this PR. Firefox is doing something really strange when doing dragover. I think this fixes that.

daee-premise:
PLEASE I NEED THIS 🙏

daee-premise:
Using react-zoom-pan-pinch with this library breaks this func completely and this fix is the only way to make both libraries work together

daee-premise:
`layerX`, `layerY` will be deprecated soon as well - this should be addressed https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/layerX

daee-premise:
@STRML

PR #1541, opened 2021-08-25
props NaN check

Arif-un:
typeof NaN === "number" .
if maxW, maxH, minW, minH get NaN value its carshing browser tab.
updated checking NaN value in propsTypes


STRML:

PR #1543, opened 2021-08-27
Replace classnames with clsx

wojtekmaj:
Follow up after https://github.com/react-grid-layout/react-draggable/pull/599

~~DO NOT MERGE YET: react-draggable needs to be updated here with version including the PR above, after that, this PR shall be rebased, which would fully remove classnames in favor of clsx, as both react-draggable and react-grid-layout depended on classnames.~~

clsx is fully compatible with classnames API, is quickly gaining traction and is about 1/3rd smaller than classnames.

example or test case.


wojtekmaj:
Updated!

Issue #463, opened 2017-01-20
Double click to drag?


arift:
- [ ] Bug
- [x] Feature Request
- [x] Question

Is there any way to make it so that the elements only drag after a double click? If not, are there any plans to implement this potential feature? 

Thank you.

PR #1607, opened 2021-11-10
Delayed drag on touch devices for better scrolling experience

Mahdi-Esbati:
The current implementation of React-Grid-Layout has a really buggy and unpredictable scrolling behavior on touch devices:

https://user-images.githubusercontent.com/17250923/141192039-4ffc5701-09ed-4df9-8ba5-d742c2b01cf2.mp4

As a result of this, I have added a "delayedDragOnTouch" prop which delays the drag behavior for 1 second, so you have to long press the grid item for 1 second before it can be dragged.
This greatly improves the scroll experience and allows the user to freely navigate around the screen:

https://user-images.githubusercontent.com/17250923/141191528-fbdc631f-e9e3-46f1-9d5c-4bfffc299978.mp4


STRML:
This is quite interesting. What do you think of the prop as a Number (ms to delay) instead of a Boolean?

Mahdi-Esbati:
> This is quite interesting. What do you think of the prop as a Number (ms to delay) instead of a Boolean?

That's a great idea!
I'll create a new commit and change it's type to a number.

Mahdi-Esbati:
@STRML
My previous approach was really buggy since I used a lot of "hacks" to handle the drag after a delay and manually calculated the "deltaX" and "deltaY" values which could sometimes lead to unpredictable behaviors.

In my refactored code, I use the 'cancel' prop in the DraggableCore component to suppress drag events.
With this new approach, we don't cancel the drag events, instead we use suppress them using the 'cancel' prop in the DraggableCore and call it's handleDragStart function directly to have a more predictable outcome.

As you suggested, I used a Number to represent the delayed drag duration and the user can set it to 0 or false to disable the delayed drag effect on touch devices.

Upon approval, I'll be more than happy to add this feature along with other highly requested features, such as double-click drag #463 [#606](https://github.com/react-grid-layout/react-draggable/issues/606), to the react-draggable/react-resizable libraries.

vinivitin:
@STRML Will this be added to the library?

Mahdi-Esbati:
@STRML
Do you have updates/suggestions about this feature?

lib/GridItem.jsx
```diff
@@ -218,6 +223,7 @@ export default class GridItem extends React.Component<Props, State> {
     // use this optimization.
     if (this.props.children !== nextProps.children) return true;
     if (this.props.droppingPosition !== nextProps.droppingPosition) return true;
+    if (this.state.allowedToDrag !== nextState.nextState) return true;
```
STRML, line 226: Seems a typo here?

lib/GridItem.jsx
```diff
@@ -415,6 +458,106 @@ export default class GridItem extends React.Component<Props, State> {
     );
   }
 
+  delayEvents = React.createRef();
+  addEvent = (type, event, passive = true) => {
```
STRML, line 462: Name could use some work; `addChildEvent` perhaps?

lib/GridItem.jsx
```diff
@@ -415,6 +458,106 @@ export default class GridItem extends React.Component<Props, State> {
     );
   }
 
+  delayEvents = React.createRef();
```
STRML, line 461: This doesn't need to be a ref, it can just be a plain array.

lib/GridItem.jsx
```diff
@@ -415,6 +458,106 @@ export default class GridItem extends React.Component<Props, State> {
...
+        this.elementRef.current.removeEventListener(type, event, {
+          passive
+        });
+      });
+      this.delayEvents.current = [];
+    }
+  };
+
+  //A reference to the timeout handler to be able to access it at any time.
+  timeoutRef: { current: null | TimeoutID, ... } = React.createRef();
```
STRML, line 483: Same, this doesn't need to be a ref.

lib/GridItem.jsx
```diff
@@ -415,6 +458,106 @@ export default class GridItem extends React.Component<Props, State> {
...
+   * Start the delayed counter to determine when a drag should start.
+   * @param  {Event} e          TouchStart/MouseDown event.
+   */
+  startDelayTimeout: Event => void = e => {
+    this.delayEvents.current = [];
+    if (!this.state.allowedToDrag) {
+      /**
+       * Register events to cancel the timeout handler if user releases the mouse or touch
+       */
+      this.addEvent("mouseup", this.resetDelayTimeout);
```
STRML, line 495: Is this needed? We should be able to handle this `onMouseUp`/`onDragStop`

lib/GridItem.jsx
```diff
@@ -415,6 +458,106 @@ export default class GridItem extends React.Component<Props, State> {
...
+      this.addEvent("touchend", this.resetDelayTimeout);
+      /**
+       * Prevent user from doing touch and scroll at the same time.
+       * If the user starts scrolling, we can not cancel the scroll event,
+       * so we cancel the drag event instead.
+       */
+      this.addEvent("touchmove", this.handleTouchMove, false);
```
STRML, line 502: Why doesn't `this.handleTouchMove()` serve this purpose?

lib/ReactGridLayoutPropTypes.js
```diff
@@ -59,6 +59,7 @@ export type Props = {|
   width: number,
   autoSize: boolean,
   cols: number,
+  dragDelayDuration: number,
```
STRML, line 62: `dragTouchDelayDuration` may be more clear

STRML (changes requested):
I like the patch's direction overall, I have some small tweaks / style questions. A few tests would really be great here as this is a lot of extra weight to carry that will easily break in future releases without tests.

lib/GridItem.jsx
Mahdi-Esbati: Oh, sorry😅!
I meant `this.state.allowedToDrag !== nextState.allowedToDrag`
I'll fix it.

lib/GridItem.jsx
Mahdi-Esbati: Ok, I'll change it to addChildEvent

lib/GridItem.jsx
Mahdi-Esbati: Fair. I'm used to creating refs for working with functional components but there's no need for them here.

lib/GridItem.jsx
Mahdi-Esbati: That's correct.

lib/GridItem.jsx
Mahdi-Esbati: There's no need for mouse events since this feature is only enabled in touch devices.

lib/GridItem.jsx
Mahdi-Esbati: This part is the most important part of delayed drag feature and cannot be removed.

We must register a non-passive touchmove event to be able to prevent scrolls when the drag delay timer has ended.

Another important note is that we must simply cancel the whole delayed drag event and reset everything if the user moves his finger before the timer is up. The reason for this is that if a user starts scrolling even for one pixel, we cannot stop the browser from scrolling afterward and e.preventDefault will not be effective.

lib/ReactGridLayoutPropTypes.js
Mahdi-Esbati: Fine, I'll change the prop name.

Mahdi-Esbati:
I've implemented the requested changes.
There was one issue on mobile devices which there was not any indication when the user was able to drag the items, so I've added a scale transform and vibration for supported devices (Safari does not support vibration API).

This is the result after these changes

https://user-images.githubusercontent.com/17250923/143223173-0cf7a0e1-b743-447c-a83c-d65c651decd0.mp4

:

Mahdi-Esbati:
I'll try to add some tests for this feature very soon.

Mahdi-Esbati:
@STRML 
This feature is almost complete.
I've tested it in the following browsers and devices and it is working flawlessly:

- [x] Google Chrome Desktop ( touch simulation using devtools )
- [x] Firefox Desktop ( touch simulation using devtools )

- [x] Google Chrome Android
- [x] Firefox Android

- [x] Safari Browser IOS 15*
- [x] Chrome Browser IOS 15*
- [x] Firefox Browser IOS 15*

* There is two small issues with the IOS 15 Browsers:
1. There is no vibration support in Safari (and other IOS browsers since they are forced to use the WebKit renderer)
2. There is a magnifying glass effect that appears on long press in IOS 15, the only way to disable it is by doing `e.preventDefault()` on `ontouchstart` event which also disables scrolling. The developers of Safari have already fixed this issue and [it is said to be released in IOS 15.2](https://bugs.webkit.org/show_bug.cgi?id=231161) and it will be disabled when the CSS `user-select` property is set to `none`

https://user-images.githubusercontent.com/17250923/143840377-0cad516e-4279-49bd-b23b-e933edc223f5.MP4

DominikScholz:
What is the state of this pull request? would be really neat to have this feature in the library.

STRML:
Taking a look at it. There was an explosion in the diff on the snapshots (was ~16k lines), have pared that down, now just making sure it works properly.

STRML:
There are also a large number of type errors that need fixing - I think we need to wire github actions into these PRs to run CI.

STRML:
Rebased on master with lint action so we can see the issues right in the PR.

Mr-sgreen:
Will this feature include in next release?It is a great feature for mobile device, hope to see this feature in next release

joshuakoh1:
Would appreciate if this gets merged for next release. Responsive grid is basically unusable on mobile devices as it hijacks normal UXUI expectations.

STRML:
Rebasing for another lint run

joshuakoh1:
Any chance this gets merged soon?

Mahdi-Esbati:
@STRML 
This merge request has been going on for more than 4 months. Please consider merging it into the project if possible. I am available to help with any issues.

STRML:
Hi @Mahdi-Esbati. Could you please fix the failing checks?

STRML:
Rebased on `master` in case that was the issue.

landa:
Any updates on this PR?

IvanShevchenko135:
Any updates on this PR?

lorenzobalzani:
Do we have news here?

mblode:
What is necessary to do in order to get this merged into main?

damnitrahul:
Any plans to merge this?

pyscukm:
@STRML @Mahdi-Esbati Any updates here? It is big problem on mobile devices to use this library and this is what we all need :)

STRML:
It needs a rebase and as my projects don't have this issue, I've not been able to spend time on it. If it were rebased and passed merge checks cleanly with concise tests, I would merge it.

PR #1613, opened 2021-11-19
onResizeStop and onDragStop callbacks should return updated layout

davidgolden:
I'm not sure why a callback for changing state would return the old state that we already have. In my opinion this should be returning the updated layout state.

### DIFF
--- a/lib/ReactGridLayout.jsx
+++ b/lib/ReactGridLayout.jsx
@@ -348,12 +348,13 @@ export default class ReactGridLayout extends React.Component<Props, State> {
       allowOverlap
     );
 
-    this.props.onDragStop(layout, oldDragItem, l, null, e, node);
-
     // Set state
     const newLayout = allowOverlap
       ? layout
       : compact(layout, compactType(this.props), cols);
+
+    this.props.onDragStop(newLayout, oldDragItem, l, null, e, node);
+
     const { oldLayout } = this.state;
     this.setState({
       activeDrag: null,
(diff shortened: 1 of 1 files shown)


STRML:
Because onLayoutChange has that new layout state. This is just to notify you of a specific action.

davidgolden:
@STRML Right, but different actions might need to be performed based on whether the grid changed from a drag or a resize, and onLayoutChange doesn't allow that.

STRML:
Hmm that makes sense. I'm thinking about why it was there. In fact, it's really not well done as the `onDragStop` handler passes a modified layout after the `moveElement` call but *before* compaction. I can't understand why this was done.

I need to look into "why" to make sure we don't break anything big but otherwise I think this is a good change. If you have a moment could you add a test?

davidgolden:
Yep I'll try to get that done today! Also one more case for this change is that `onLayoutChange` is called on initial render, which isn't suitable for change effects.

davidgolden:
@STRML This one got away from me! I'm trying to add a test and struggling to figure out how to simulate a drag or resize event such that the callbacks are called. Do you have any pointers? I know this issue has been a problem for others as well, so would be great to get it out there.

STRML:
There should be some examples of simulating that in the react-draggable and react-resizable libraries that will help out. You could just copy the helpers over.

davidgolden:
Right that is the first place I looked, but those don't seem to trigger the callbacks for whatever reason. I've tried a few other things but nothing's worked for me so far.

davidgolden:
Should not be stale. Without these changes `onResizeStop` and `onDragStop` are pretty much useless.

mcle0463:
I've found that onLayoutChange is called during the initial render as well (no layout changes), which means that I cant rely on this function to save an updated state. I will need to compare the args to what is currently in state and then save. I tried to get around this by using onResizeStop and onDragStop, which would allow me to save the updated state when one of these events occur to guarantee an update event. They seem to trigger as expected, but they dont pass an object array of allLayouts like onLayoutChange handler does, which renders them useless for my case. Maybe i'm not understanding how they are meant to be used.

PR #1619, opened 2021-11-26
deeply clone the layout into oldLayout to ensure when we compare the …

### LINKED ISSUE #1606: When using allowOverlap prop the onLayoutsChanged event is not fired when dragging of an item ends
Issue: When using allowOverlap prop the onLayoutsChanged event is not fired onDragStop. 

I have made a sandbox showing the issue. Watch the counters above the grid and move the items to see the issue when items are dragged. This is not affecting resizing of items.

- Expected behaviour: 
- Each time and item is dragged: 
-      onDragStop counter increases by 1 ✅
-      onLayoutChanged counter increases by 1 ❌
-      onResizeStop Counter remains the same ✅
-   Each time an item is resized: 
-      onResizeStop counter is increased by 1 ✅
-      onLayoutChanged counter is increased by 1 ✅
-      onDragStop counter remains the same ✅

Example of issue: https://rb7fq.csb.app/
Sandbox of issue: https://codesandbox.io/s/exciting-brahmagupta-rb7fq

woollsta:
…oldLayout it correctly detects it as changed (fixes #1606)

example or test case.

### DIFF
--- a/lib/ReactGridLayout.jsx
+++ b/lib/ReactGridLayout.jsx
@@ -2,6 +2,7 @@
 import * as React from "react";
 
 import isEqual from "lodash.isequal";
+import cloneDeep from "lodash.clonedeep";
 import clsx from "clsx";
 import {
   bottom,
@@ -56,6 +57,7 @@ type State = {
 };
 
 import type { Props, DefaultProps } from "./ReactGridLayoutPropTypes";
+
 
 // End Types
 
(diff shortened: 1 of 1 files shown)


STRML:
What issue does this solve? And why not use the `cloneLayout` helper which is much faster and doesn't require a dependency?

STRML:
Ah I see. The layout is directly mutated when `allowOverlap` is present in #1619.

woollsta:
Thanks, good spot, I'm not too familar with your project. Using the cloneLayout now and still working.

STRML:
Fixed in #1620, thanks for bringing this to my attention.

PR #1643, opened 2022-01-07
more GridItem tests

imagineLife:
Increase code coverage of `<GridItem/>`


lib/GridItem.jsx
```diff
@@ -129,28 +129,24 @@ export default class GridItem extends React.Component<Props, State> {
     // All optional
     minW: function (props: Props, propName: string) {
       const value = props[propName];
-      if (typeof value !== "number") return new Error("minWidth not Number");
```
STRML, line 132: Why get rid of these?

imagineLife: Hey @STRML I'm shooting for increasing the unit-test coverage here - this code does not seem testable by passing minW as a string to trigger this error - [check out this previous commit](https://github.com/react-grid-layout/react-grid-layout/runs/4734415506?check_suite_focus=true) that I tried out where I passed minWidth as a string && the prop-types "caught" the prop type error

Seems like there could be a few approaches, maybe?
1 - leave this code in the codebase and comment/note that these lines will not be "testable" in unit tests because the prop-types aren't currently allowing passing bad types  
2 - (_like here_) remove this js check in favor of removing non-testable code
3 - some other approach? where this can be tested _and_ left in the code

STRML: Oh I see. Yes, Flow caught that. A very common thing to do in tests is to add an ignore comment, like this, because we can't assume people using the library are actually using Flow or TypeScript.

https://github.com/react-grid-layout/react-grid-layout/blob/388b79fb52fabfe107f946fbad9ffacd63324d09/test/spec/lifecycle-test.js#L99

The comment will suppress the error. Run `npm run flow` or `yarn flow` to ensure it works before committing.

imagineLife: ah 👌 will update ✅   
[Its a small world, I suppose](https://github.com/facebook/flow/issues/8538), flow was hanging on my m1 - I just added a `sharedmemory.heap_size=3221225472` for local workaround

STRML: It actually shouldn't take that workaround anymore if you're up to date on Flow and OSX versions.

STRML:
Thanks. Want me to review it again @imagineLife

imagineLife:
hey @STRML sure!

STRML:
Yeah - great. Thanks!

PR #1645, opened 2022-01-13
Add GS to 'Projects using React-Grid-Layout' section, README.md

duncanhope:
Add Goldman Sachs to 'Projects using React-Grid-Layout' section

example or test case.


STRML:
Thanks!

There is some pretty high-quality documentation on that page; are you able to get permission from your end for us to use the graphics in the README here? It could improve the experience for all users of RGL.

duncanhope:
Hi Sam,

I don't actually work at Goldmans - I was also looking at their design
guidelines as I'm at a different bank but also doing react dev work and as
you've noticed, it's got some useful docs.
Was reassured to see they use it (and it prompted me to start looking at
using it too)

So I couldn't provide any permission as it's not my content.

PR #1704, opened 2022-04-12
Fix compactItem for case when allowOverlap is true and compactType is null

### LINKED ISSUE #1649: dragged from outside and allowOverlap=true
If dragged from outside into react-gird-layout, it will trigger reordering, when I set compactType=null, allowOverlap=true, isDroppable=true

benedict-jackson:
Fixes #1649 

Avoids entering collision loop when allowOverlap is true and compactType is null.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -316,7 +316,7 @@ export function compactItem(
 
   // Move it down, and keep moving it down if it's colliding.
   let collides;
-  while ((collides = getFirstCollision(compareWith, l))) {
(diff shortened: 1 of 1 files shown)


Vokabelsalat:
Would be interested in the solution, because allowOverlap isn't working together with drop from outside in a Responsive Layout.

STRML:
Closing: This feature/fix is now included in the v2 TypeScript rewrite. The functionality has been incorporated into the core architecture. Thank you for the contribution!

PR #1705, opened 2022-04-13
Avoid unwanted re-render on resize

dubzzz:
While digging into the source code of the library to investigate a runtime crash I had in one of my app, I found this quick optimization. I though it could still be useful to share it but feel free to reject it if it does not make sense.

---

Up-to-now the current implementation watches the event `onWindowResize` and triggers a state update anytime it receives an event related to resize even if the size of the offset was not updated.

The PR just removes this unneeded re-render by not updating the state in such case.

### DIFF
--- a/lib/components/WidthProvider.jsx
+++ b/lib/components/WidthProvider.jsx
@@ -79,7 +79,9 @@ export default function WidthProvideRGL<Config>(
       // fix: grid position error when node or parentNode display is none by window resize
       // #924 #1084
       if (node instanceof HTMLElement && node.offsetWidth) {
-        this.setState({ width: node.offsetWidth });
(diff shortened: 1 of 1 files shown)


STRML:
Thanks. It seems fairly unlikely that the window would resize and the grid *wouldn't* change width, unless of course it was being used in a fixed-width layout. But if it were, why would you be using `<WidthProvider>` at all?

dubzzz:
Trying to find back why I needed it 😂 Will come back to you when I get a better idea of my original need (don't remember at all why I proposed the fix)

dubzzz:
Just falling back on this PR after a few years. As it seems to be a free optimization, is there any reason not to merge it. Or if it's never going to be merged maybe it's worth closing it.

Issue #1638, opened 2021-12-24
【bug】When I use onlayoutchange, onlayoutchange will not be triggered during dragging.


jiaicon:
Look!
This is what caused the problem.
![image](https://user-images.githubusercontent.com/34124517/147308553-0e0b8e39-6f43-4e1e-be0b-f176350b9f48.png)

hornta:
What problem? What are you reffering to?

livandrey:
When  `allowOverlap = true` and moving without collisions onLayoutChange will not be triggered

To fix this:
lib/ReactGridLayout.jsx:

in `onDragStart`:
`oldLayout: layout` -> `oldLayout: cloneLayout(layout)`

same in `onResizeStart`:
`oldLayout: this.state.layout` -> `oldLayout: cloneLayout(this.state.layout)`

artemlaptev63:
> When `allowOverlap = true` and moving without collisions onLayoutChange will not be triggered
> 
> To fix this: lib/ReactGridLayout.jsx:
> 
> in `onDragStart`: `oldLayout: layout` -> `oldLayout: cloneLayout(layout)`
> 
> same in `onResizeStart`: `oldLayout: this.state.layout` -> `oldLayout: cloneLayout(this.state.layout)`

have the same issue, the solution did not help

artemlaptev63:
> When `allowOverlap = true` and moving without collisions onLayoutChange will not be triggered
> 
> To fix this: lib/ReactGridLayout.jsx:
> 
> in `onDragStart`: `oldLayout: layout` -> `oldLayout: cloneLayout(layout)`
> 
> same in `onResizeStart`: `oldLayout: this.state.layout` -> `oldLayout: cloneLayout(this.state.layout)`

@livandrey have the same issue, the solution did not help

R4DIC4L:
For anyone else trying this function, I had to specifically add allowOverlap={false} property, it is probably set on true by default.

chandlervdw:
You can add the same `ItemCallback` to `onLayoutChange`, `onDragStop` and `onResizeStop` in order to update the layout on drag/resize when `allowOverlap` is true.

PR #1709, opened 2022-04-18
On drag/resize, fire `onLayoutChange` when `allowOverlap` is true

chandlervdw:
When `allowOverlap` is true, `onLayoutChange` is not fired unless there is collision, as per #1638. This PR follows @livandrey's suggestion, ensuring that a clone of the state's layout is always passed to state so that dragging and resizing will always subsequently fire `onLayoutChange`.

### DIFF
--- a/lib/ReactGridLayout.jsx
+++ b/lib/ReactGridLayout.jsx
@@ -6,6 +6,7 @@ import clsx from "clsx";
 import {
   bottom,
   childrenEqual,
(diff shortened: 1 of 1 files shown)


chandlervdw:
This is unnecessary, as I can pass the same `ItemCallback` to `onDragStop` and `onResizeStop` as I do to `onLayoutChange` and I can update the layout.

PR #1734, opened 2022-06-15
Add get layout

ankero:
Adds getLayout function that can be accessed through the ReactGridLayout dom instance (through ref.current.getLayout()). 

This makes it possible to query the layout state after modifying the grid and letting it fix potential overlapping elements. Without this function we cannot get the correct state without inducing a change that propagates to onLayoutChange, which doesn't seem to be possible.

Additionally this allows accessing the corrected state if allowOverlap={false} and preventCollision={false} and an element is moved or resized to push another element(s) out of the way. In these cases the onLayoutChange does not return the correct value for the pushed elements.

Usecases:

* Insert a new element on top of a grid that already has multiple elements. This pushes other elements away yet does not fire onLayoutChange with updated values
* Verify layout after onLayoutChange callback is fired as that function does not return correct values for elements that have been pushed to new location
* Verify layout after element moved, sometimes it seems that the onLayoutChange state does not return the correct x -value if the grid collapses vertically & user tried to move the element to a higher x value than it was collapsed to.

Notes:
The correct values should be fixed in the onLayoutChange and it should fire always when the grid changes and only with the final values. However this getLayout -function will "fix" the issue by creating a workaround.


ankero:
Never mind, found a way to access the state directly by using the proper ref 🙄 (I was trying to use the innerRef and not ref property)

For someone searching for solution for this issue:
in parent:
`
const ref = useRef();

const onLayoutChange = () => {
console.log(ref.current.state.layout);
}

return <ReactGridLayout ref={ref} />
`

PR #1749, opened 2022-07-06
Disabled useCSSTransforms by default on Firefox due to Mozilla bug wi…

### LINKED ISSUE #1748: useCSSTransforms is broken in Firefox with SVGs due to Mozilla bug
### Describe the bug

useCSSTransforms is enabled by default in this library but mouse events are incorrectly offset on SVGs inside a React Grid Item on Firefox.

In my use case, I was displaying graphs inside React Grid Items, but no mouse cursor events were firing inside the graphs due to large CSS3 transforms being applied to the parent Grid Item element.

This is due to a bug in Firefox re: "SVG getScreenCTM does not return correct position if parent is transformed". See https://bugzilla.mozilla.org/show_bug.cgi?id=1610093 for more information, and http://jsfiddle.net/edemaine/vLjd1pa7/6/ for a fiddle that demonstrates the erroneous behavior in Firefox vs Chrome.  

N/A

See bug description.

### Expected behavior

Consistent behavior of useCSSTransforms on Firefox and Chrome. 

^1.3.4

### Operating System Version

macOS

### Browser

Firefox

### Additional context

_No response_

### Screenshots or Videos

_No response_

BBarash:
…th SVG and CSS3 transforms. Fixes #1748.

### DIFF
--- a/README.md
+++ b/README.md
@@ -332,6 +332,8 @@ isResizable: ?boolean = true,
 isBounded: ?boolean = false,
 // Uses CSS3 translate() instead of position top/left.
 // This makes about 6x faster paint performance
+// Disabled on Firefox by default due to negative SVG interaction.
+// See https://bugzilla.mozilla.org/show_bug.cgi?id=1610093
 useCSSTransforms: ?boolean = true,
 // If parent DOM node of ResponsiveReactGridLayout or ReactGridLayout has "transform: scale(n)" css property,
 // we should set scale coefficient to avoid render artefacts while dragging.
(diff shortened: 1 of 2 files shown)


STRML:
This is too blunt a hammer. I would suggest a development warning instead if an svg child is detected.

STRML:
Closing this PR - v2 provides a cleaner solution via the `PositionStrategy` interface.

If you're rendering SVG content and experiencing Firefox issues, you can use `absoluteStrategy` instead of the default transform-based positioning:

```tsx
import { absoluteStrategy } from 'react-grid-layout';

<GridLayout positionStrategy={absoluteStrategy} ... />
```

This avoids browser sniffing and gives users explicit control. Thanks for the contribution!

Issue #665, opened 2017-11-26
Store additional layout data


rakeen:
Hi, great component you made here. kudos :+1: 

Please mark the type of this issue:

- [ ] Bug
- [x] Feature Request
- [x] Question

I was wondering if you have any plans to store additional data inside react-grid-layout?

By additional I mean some extra fields in the layout object. e.g.
`{i: 'c', x: 4, y: 0, w: 1, h: 2, data: 'myAdditionalData'}`

Currently if we add additional data, it gets removed when the layout is changed.
```js
onLayoutChange = (layout) => {
    // layout doesn't have data
}
```

One workaround is to append the extra data on **every** `onLayoutChange` event!

codepunkt:
One workaround is to have an array or dictionary with a unique identifier string that you use as each layout object's `i`. Store the data somewhere else and reference it.

Asif-Saheer-k:
const onLayoutChange = (oldLayout, newLayout) => {
    const updateLayout = newLayout.map((item, key) => ({
      ...item,
      name: oldLayout[key]?.name,
      display: oldLayout[key]?.display
    }));
    setLayout(updateLayout);
    if (!block) {
      setPreLayout(prevLayouts => [...prevLayouts, newLayout]);
    } else {
      setBlock(false);
    }
  };
  
  
     onLayoutChange={newItem => onLayoutChange(layout, newItem)}
     
     
     above problem i handled this way

PR #1776, opened 2022-09-18
add ability to pass additional info to layout items

k1og:
Add ability to provide addinional information to LayoutItem #665

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -134,6 +134,7 @@ export function withLayoutItem(
 // Fast path to cloning, since this is monomorphic
 export function cloneLayoutItem(layoutItem: LayoutItem): LayoutItem {
   return {
+    ...layoutItem,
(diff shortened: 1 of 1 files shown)


k1og:
currently doesn't work with data-grid attribute, so WIP

k1og:
nvm
i was wrong. it actually does work with both data-grid attribute and layout prop

STRML:
Closing this PR - we've decided not to support arbitrary additional properties on LayoutItem.

If you need to associate custom data with grid items, we recommend managing it separately using the item's `i` (id) as a key:

```tsx
const itemData = {
  'item-1': { customProp: 'value' },
  'item-2': { customProp: 'other' }
};

// Access via: itemData[item.i]
```

This keeps the layout data structure clean and predictable. Thanks for the contribution!

PR #1777, opened 2022-09-20
feat: add lockAspectRatio props in GridItem Component

Kun8018:

example or test case.


STRML:
Closing - aspect ratio locking can be implemented via the `onResize` callback without adding complexity to the core library:

```tsx
onResize={(layout, oldItem, newItem, placeholder) => {
  const ratio = 16/9; // or item-specific ratio
  placeholder.h = newItem.h = Math.round(newItem.w / ratio);
}}
```

This gives you full control over the aspect ratio logic per-item. Thanks for the contribution!

PR #1781, opened 2022-10-06
Avoid possible stack overflows into `resolveCompactionCollision`

dubzzz:
While running `react-grid-layout` on one of my project, I encountered a case leading `resolveCompactionCollision` into a stack overflow. The reason being that the cursion had to apply on too many colliding items and thus led us to a runtime crash.

The current PR changes the implementation of the algorithm into a non recursive version by replacing the native stack handled by the browser (and pretty limited in size) my a manually handled one backed by an array and behaving exactly the same way.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -252,35 +252,41 @@ function resolveCompactionCollision(
   moveToCoord: number,
   axis: "x" | "y"
 ) {
-  const sizeProp = heightWidth[axis];
-  item[axis] += 1;
-  const itemIndex = layout
-    .map(layoutItem => {
-      return layoutItem.i;
-    })
-    .indexOf(item.i);
+  // Instead of performing recursive calls to the function itself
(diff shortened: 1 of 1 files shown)


STRML:
How many collisions did you have?

Any performance difference? I would expect this will actually be quite a bit faster.

dubzzz:
I'll do some extra measurements and come back to you

dubzzz:
@STRML I think I got the reason of my problem. I got a layout with 177 items but most of them have been **duplicated in terms of identifiers**. Given the figures below, I think we can probably not merge this PR as the issue can only happen for layouts having like 200 collisions which does not seem that reasonable 🤔 

More precisely: I had like 89 unique identifiers making my whole layout, it made the code dealing with collisions somehow running deeper than it should have run making it explodes with depth of 199 calls to `resolveCompactionCollision`.

PR #1782, opened 2022-10-12
Add checking the compactType null value to avoid breaking the layout when overlapping is allowed.

jaloveeye:
When allowOverlap is true and compactType is null, the layout is unintentionally transformed like the AS-IS gif.

AS-IS

Adding an exception when compactType is null in compactItem avoids unintended layout transformed.

PR


STRML:
Shouldn't this also be checking if `allowOverlap` is true? Otherwise, when it is the default value, no collisions will occur when compactType is null.

jaloveeye:
@STRML 
Thank you for your advice.
I was missing out that.
Your advice has been reflected.

STRML:
Thanks!

PR #1801, opened 2022-11-21
feat: Add a stream interception to the onDragOver event

githubobil:
The Question：
 Calls when an element has been dropped into the grid from outside，The trigger frequency of ondragover events is too high, causing the problem of stuck
image

image

The Solution:
ondragover adds interception

The Effect
image

### DIFF
--- a/lib/ReactGridLayout.jsx
+++ b/lib/ReactGridLayout.jsx
@@ -2,6 +2,7 @@
 import * as React from "react";
 
 import isEqual from "lodash.isequal";
+import throttle from "lodash.throttle"
 import clsx from "clsx";
 import {
   bottom,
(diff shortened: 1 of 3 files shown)


STRML:
Please don't commit package.json changes and built files.

avasuro:
Please, keep in mind that inside `onDragOver` handler end user could  do some actions to affect dnd process, e.g. apply `e.dataTransfer.dropEffect`, which affects visual look of mouse pointer, ability to dop (e.g. prevents drop if set to 'none') and could be used later in onDrop handler.

If onDragOver is throttled - dropEffect will be applied not immediately and this could produce bugs in end-user applications.

So, in my opinion only re-render of grid layout could be trottled to improve performance, but all DnD-related staff (like call of `onDragOver` handler etc.) should be handled immediately. In other way - there is very high risk of various DnD bugs for the end users.

STRML:
Closing - v2 has rewritten the external drop handling with hooks.

If you're experiencing performance issues with drag-over events in v2, please open a new issue with a reproduction. The throttling approach is valid but we'd want to test it with the new implementation first.

Thanks for the contribution!

PR #1822, opened 2023-01-04
fix(grid): Fixed Compact horizontal not always moving elements as left as possible.

llorenspujol:
Compact "horizontal" is not moving elements as left as possible when some item overflows to the right. It just moves items down but not to the left.

Take for example this layout:

```typescript
const layout = [
      {y: 0, x: 0, h: 2, w: 2, i: '1'},
      {y: 0, x: 2, h: 2, w: 2, i: '2'},
      {y: 0, x: 4, h: 2, w: 2, i: '3'},
      {y: -2, x: -2, h: 2, w: 2, i: '4'}
    ];
```

where the result after compacting (WITHOUT THE FIX) is:
```typescript
   expect(compact(layout, 'horizontal', cols)).toEqual([
      {y: 0, x: 2, h: 2, w: 2, i: '1', moved: false, static: false},
      {y: 0, x: 4, h: 2, w: 2, i: '2', moved: false, static: false},
      {y: 2, x: 4, h: 2, w: 2, i: '3', moved: false, static: false},
      {y: 0, x: 0, h: 2, w: 2, i: '4', moved: false, static: false},
   ]);
```

As we see, the element with id = 3 is not moved to the left, when it should. It is also a smell, that if we apply the same compact "horizontal" function again, we will have the correct layout. Executing the compact function more than 1 time on the same layout should always return the same result, not a different one.

The expected result should be:
```typescript
   expect(compact(layout, 'horizontal', cols)).toEqual([
      {y: 0, x: 2, h: 2, w: 2, i: '1', moved: false, static: false},
      {y: 0, x: 4, h: 2, w: 2, i: '2', moved: false, static: false},
      {y: 2, x: 0, h: 2, w: 2, i: '3', moved: false, static: false},
      {y: 0, x: 0, h: 2, w: 2, i: '4', moved: false, static: false},
   ]);
```

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -326,6 +326,10 @@ export function compactItem(
     if (compactH && l.x + l.w > cols) {
       l.x = cols - l.w;
       l.y++;
+      // ALso move element as left as we can
+      while (l.x > 0 && !getFirstCollision(compareWith, l)) {
+        l.x--;
+      }
     }
   }
 
--- a/test/spec/utils-test.js
+++ b/test/spec/utils-test.js
@@ -493,6 +493,23 @@ describe("compact horizontal", () => {
       { y: 5, x: 2, h: 1, w: 1, i: "5", moved: false, static: true }
     ]);
   });
+
+  it('Should put overflowing right elements as bottom needed without colliding and as left as possible', () => {
+    const cols = 6;
+    const layout = [
+      {y: 0, x: 0, h: 2, w: 2, i: '1'},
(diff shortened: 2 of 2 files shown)


STRML:
Thanks!

llorenspujol:
You are welcome! 

I ported this lib to [angular](https://github.com/katoid/angular-grid-layout), and I used most of the `lib/utils.js` for it... So now I am giving back a little 😆

 Great work! 👏

PR #1844, opened 2023-02-14
fix[synchronizeLayoutWithChildren]:  When synchronizing layouts, upda…

MichelJay:
…te pre-existing layouts rather than clone them forever

example or test case.


STRML:
It's not intended that `data-grid` properties should be continuously mixed into the layout. But I'm open to understanding why you made thsi change.

PR #1847, opened 2023-02-24
feat(audit): add packj

STRML:
Trial run of https://github.com/ossillate-inc/packj

### DIFF
--- a/.github/workflows/packj.yml
+++ b/.github/workflows/packj.yml
@@ -1,0 +1,5 @@
+- name: Packj Security Audit
+  uses: ossillate-inc/packj-github-action@0.0.6-beta
+  with:
+    DEPENDENCY_FILES: npm:package.json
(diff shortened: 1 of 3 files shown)


ADejbakhsh:
Hello, i was wondering if you managed to make packj work.
I looked at your file and there is a few thing missing and i was wondering if it was intentional ?
I can't find any trigger like this
```YAML
# Controls when the workflow will run
on:
  pull_request:
    branches:
      - master
```
Nor can i see the required arguement: runs-on:
```YAML
    runs-on: ubuntu-latest
```
*Note it can run only on linux according to the README

You are also missing the [.packj.yaml](https://github.com/ossillate-inc/packj/blob/main/.packj.yaml) file, the job won't even run without it on my side.

I just want confirmation that the documentation is not ready, and that i am not going crazy.

STRML:
Nah, you're right. I got about 80% through this, found it a bit hard to test, then got distracted with the kids. I'll try to finish it up today.

PR #1853, opened 2023-03-05
Replace `lodash.isequal` to `fast-equals`

Semigradsky:
Suggestion to replace `lodash.isequal` to well-maintained faster implementation.


STRML:
Yeah, I'm into it. Thanks.

PR #1899, opened 2023-06-22
Expose enableUserSelectHack prop for Draggable

kguyton6:

example or test case.

### DIFF
--- a/lib/GridItem.jsx
+++ b/lib/GridItem.jsx
@@ -70,6 +70,7 @@ type Props = {
   // Draggability
   cancel: string,
   handle: string,
+  enableUserSelectHack?: boolean,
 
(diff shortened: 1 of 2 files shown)


n1ghtmare:
Hi, thanks for the PR, could you please add some description to it? Add some context.

STRML:
Closing - this can be handled with CSS in v2:

```css
/* Prevent text selection during drag */
.react-grid-item { user-select: none; }

/* Or only during active drag */
.react-draggable-dragging { user-select: none; }
```

If there's strong demand for fine-grained control of react-draggable options, we could add them to `dragConfig` in the future, but CSS handles most use cases.

Thanks for the contribution!

PR #1915, opened 2023-08-04
Fix position logic when item is dragged in from the outside

AntSan813:
This fixes an issue where if you have a nested grid, the position of an element dragged in from the outside will not be computed correctly. 

As a plus, this also moves away from using the deprecated `layerX` and `layerY` event properties.


STRML:
Thank you, very nice.

PR #1918, opened 2023-08-16
Export calculateUtils

dseif:
## Goal
export `calculateUtils`.

## Reasoning
We've found this utils useful when writing some custom `onResize` and `onDrag` implementations. Currently we just copied the utils to reuse them, but figured it'd make more sense to just export them (and likely useful to others as well).

This also helps ensure these calculations are consistent both within the library and for any external consumers looking to augment the default behaviour.

## Other considerations
One potential downside is that many of these utils require consumers to pass `positionParams` as an argument. While this isn't bad (the user has all this information), it does feel like an extra step. I wonder if there is a way to expose something like `getPositionParams` from `GridItem` to make this cleaner?

However, something like that would require exposing a RGL instance and having a method hung off that (since all the positional information is relative to the associated grid). Probably outside of the scope of this PR, but something I figured I'd mention just incase you have any thoughts.

### DIFF
--- a/index-dev.js
+++ b/index-dev.js
@@ -1,5 +1,6 @@
 module.exports = require("./lib/ReactGridLayout").default;
 module.exports.utils = require("./lib/utils");
+module.exports.calculateUtils = require("./lib/calculateUtils");
 module.exports.Responsive = require("./lib/ResponsiveReactGridLayout").default;
 module.exports.Responsive.utils = require("./lib/responsiveUtils");
 module.exports.WidthProvider =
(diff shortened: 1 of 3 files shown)


STRML:
I agree with your note, a helper would be useful.

Also, it's a bit awkward to access `utils.calculateUtils`. Why not export it from `index.js`?

dseif:
> Also, it's a bit awkward to access utils.calculateUtils. Why not export it from index.js?

Fair enough, happy to just export it from `index.js`. My reasoning was trying to keep the utils grouped and accessible under the same property.

dseif:
re: the helper - Like I mentioned going to leave that out of this PR. Once I get the other I have in flight done I'll circle back to figuring out how to expose it.

STRML:
Nope! :)

PR #1923, opened 2023-08-25
fix: set activeDrag in the state onDragStart

tantawyk:
this fixes a problem where clicking on the drag handle triggers onDragStart and doesn't trigger onDragStop as the activeDrag was being set only onDrag and not onDragStart as well. another solution would be to remove the check for activeDrag in the onDragStop function.

https://github.com/react-grid-layout/react-grid-layout/issues/1745

### DIFF
--- a/lib/ReactGridLayout.jsx
+++ b/lib/ReactGridLayout.jsx
@@ -249,10 +249,21 @@ export default class ReactGridLayout extends React.Component<Props, State> {
     const { layout } = this.state;
     const l = getLayoutItem(layout, i);
     if (!l) return;
+    
+    // Create placeholder (display only)
+    const placeholder = {
+      w: l.w,
+      h: l.h,
+      x: l.x,
+      y: l.y,
+      placeholder: true,
+      i: i
+    };
 
     this.setState({
       oldDragItem: cloneLayoutItem(l),
-      oldLayout: layout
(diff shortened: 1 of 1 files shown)


tantawyk:
@STRML please let me know if this is not wanted, or if it needs some tests

STRML:
This was years ago, so I'm trying to figure out the intent - I think it was that it didn't make much sense to show the placeholder only on `onDragStart` in case it was a simple click with no movement.

It is definitely confusing that `onDragStart` can fire with no `onDragStop`, so perhaps it is better to fix that issue.

STRML:
Went ahead and merged this to try to get on top of React 18 issues.

cyuan01:
I know this is from a while ago but just noting @STRML and @tantawyk I believe this change makes it impossible to have any click behavior within the grid items themselves. For example, in this [codesandbox](https://codesandbox.io/p/sandbox/awesome-kare-tj7dy4?file=%2Fpackage.json%3A14%2C23) the "button" within the grid item can't be clicked properly because the onDragStart is for lack of better words, too eagerly setting activeDrag and I think preventing other click events from happening. 

I narrowed things down by confirming click works in 1.4.2 release but not 1.4.3 release, which only has this change in it. Thoughts?

PR #1930, opened 2023-09-12
Allow up to top=0 when resizing items. Fixes #1929

### LINKED ISSUE #1929: [v1.4.0] [Regression] Cannot change height of items in the first row when containerPadding is [0, 0]
### Describe the bug

Cannot change height of items in the first row when containerPadding is [0, 0]. Setting containerPadding to [1, 1] fixes the issue.

https://codesandbox.io/s/peaceful-dewdney-4y9gx2?file=/src/ShowcaseLayout.js

Just use the resize handle on an item of the first row, you can't change the height.

### Expected behavior

I'm able to resize width AND height of items of the first row

1.4.0

### Operating System Version

Linux

### Browser

Chrome

### Additional context

_No response_

### Screenshots or Videos

_No response_

ddolcimascolo:

example or test case.

### DIFF
--- a/lib/GridItem.jsx
+++ b/lib/GridItem.jsx
@@ -578,25 +578,6 @@ export default class GridItem extends React.Component<Props, State> {
   ): void {
     const handler = this.props[handlerName];
     if (!handler) return;
-    const { x, y, i, maxH, minH } = this.props;
-    let { minW, maxW } = this.props;
-
-    // Get new XY
-    let { w, h } = calcWH(
-      this.getPositionParams(),
-      size.width,
-      size.height,
-      x,
-      y,
-      handle
-    );
-
-    // minW should be at least 1 (TODO propTypes validation?)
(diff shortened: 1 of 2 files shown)


lib/GridItem.jsx
```diff
@@ -676,6 +656,26 @@ export default class GridItem extends React.Component<Props, State> {
       });
     }
 
+    const { x, y, i, maxH, minH } = this.props;
```
ddolcimascolo, line 659: I had to move this part _after_ the updated size is computed, and used the updated size to compute the new w and h coordinates. It seems more logical like that but I honestly don't know the library enough to make sure this change is OK... 

Thoughts @STRML  ?

STRML: Yes this is a good change - thanks!

STRML:
Fixed in #1931 with your changes, thank you!

PR #1917, opened 2023-08-15
Multi directional resizing

dseif:
## Goal
Make resizing from all supported directions function like the current E/S/SE do.

## Details
This is almost entirely taken from the great work done in this PR https://github.com/react-grid-layout/react-grid-layout/pull/1377 by [ludovic](https://github.com/ludovic) (I don't know how to ping/tag him or else I would). I'd like to make sure all his work is acknowledged here so I made sure everything he did has him as the commit author. If this gets squash merged can we make sure this is co-authored by him?

The changes I made were to address some of the comments that @STRML had in the two videos he posted. Resizing should now be properly constrained to the container and all sides but the bottom.

I plan on cleaning up and consolidating some of this code to make it all a bit neater, but wanted to get this up as early as I could to solicit any other feedback you may have.

## Video of the change in action
https://github.com/react-grid-layout/react-grid-layout/assets/580086/13dc529a-5ae1-4118-91aa-4f9b0e4d2f63

PR #1937, opened 2023-09-14
Resizing now obeys preventCollision restrictions

### LINKED ISSUE #1933: preventCollision not taking effect while resizing items
### Describe the bug

After 1.4.0 version, preventCollision appears to be being ignored when resizing items.

https://codesandbox.io/s/elated-napier-gr5lgl

1. Change Compaction Type to No Compaction
2. Use resize handle in horizontal (element itself will be pushed down until find some free space)
3. Use resize handle in vertical (elements bellow will be pushed)

### Expected behavior

When resizing with preventCollision, I expect the element to just fill the free space when available without moving out of place and not pushing the other elements, exactly the same way as in version 1.3.4

1.4.1

### Operating System Version

Linux

### Browser

Chrome

### Additional context

_No response_

### Screenshots or Videos

_No response_

dseif:
## Description

This PR fixes #1933 

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [x] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Related Tickets & Documents

This bug was introduced as part of the work in #1917 because `preventCollision` wasn't taken into account when performing resizes.

## Mobile & Desktop Screenshots/Recordings
Both examples have `preventCollision=true` and `compactType=null` as was defined in #1933.

### Before
https://github.com/react-grid-layout/react-grid-layout/assets/580086/2f24816d-68ff-488e-ad1c-c4e2b0f9af8e

### After
https://github.com/react-grid-layout/react-grid-layout/assets/580086/12dd558f-8af3-4a7d-9ca1-4559fde81da9

## Added tests?

- [x] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 examples

### DIFF
--- a/lib/ReactGridLayout.jsx
+++ b/lib/ReactGridLayout.jsx
@@ -405,18 +405,50 @@ export default class ReactGridLayout extends React.Component<Props, State> {
     let { layout } = this.state;
     const { cols, preventCollision, allowOverlap } = this.props;
 
+    let shouldMoveItem = false;
+    let finalLayout;
+    let x;
+    let y;
+
     const [newLayout, l] = withLayoutItem(layout, i, l => {
+      let hasCollisions;
+      x = l.x;
+      y = l.y;
+      if (["sw", "w", "nw", "n", "ne"].indexOf(handle) !== -1) {
+        if (["sw", "nw", "w"].indexOf(handle) !== -1) {
+          x = l.x + (l.w - w);
+          w = l.x !== x && x < 0 ? l.w : w;
+          x = x < 0 ? 0 : x;
+        }
+
+        if (["ne", "n", "nw"].indexOf(handle) !== -1) {
+          y = l.y + (l.h - h);
+          h = l.y !== y && y < 0 ? l.h : h;
+          y = y < 0 ? 0 : y;
+        }
+
+        shouldMoveItem = true;
+      }
+
       // Something like quad tree should be used
       // to find collisions faster
-      let hasCollisions;
       if (preventCollision && !allowOverlap) {
-        const collisions = getAllCollisions(layout, { ...l, w, h }).filter(
-          layoutItem => layoutItem.i !== l.i
-        );
+        const collisions = getAllCollisions(layout, {
+          ...l,
+          w,
+          h,
+          x,
+          y
+        }).filter(layoutItem => layoutItem.i !== l.i);
         hasCollisions = collisions.length > 0;
 
         // If we're colliding, we need adjust the placeholder.
         if (hasCollisions) {
+          // Reset layoutItem dimensions if there were collisions
+          y = l.y;
+          h = l.h;
+          x = l.x;
+          w = l.w;
           // adjust w && h to maximum allowed space
           let leastX = Infinity,
             leastY = Infinity;
@@ -429,6 +461,8 @@ export default class ReactGridLayout extends React.Component<Props, State> {
           if (Number.isFinite(leastY)) l.h = leastY - l.y;
         }
       }
+      l.w = w;
+      l.h = h;
 
       return l;
     });
@@ -436,24 +470,8 @@ export default class ReactGridLayout extends React.Component<Props, State> {
     // Shouldn't ever happen, but typechecking makes it necessary
     if (!l) return;
 
-    let finalLayout;
-    if (["sw", "w", "nw", "n", "ne"].indexOf(handle) !== -1) {
-      let x = l.x;
-      let y = l.y;
-      if (["sw", "nw", "w"].indexOf(handle) !== -1) {
-        x = l.x + (l.w - w);
-        w = l.x !== x && x < 0 ? l.w : w;
-        x = x < 0 ? 0 : x;
-      }
-
-      if (["ne", "n", "nw"].indexOf(handle) !== -1) {
-        y = l.y + (l.h - h);
-        h = l.y !== y && y < 0 ? l.h : h;
-        y = y < 0 ? 0 : y;
-      }
-
-      l.w = w;
-      l.h = h;
+    finalLayout = newLayout;
+    if (shouldMoveItem) {
       // Move the element to the new position.
       const isUserAction = true;
       finalLayout = moveElement(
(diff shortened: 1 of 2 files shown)


lib/ReactGridLayout.jsx
```diff
@@ -429,31 +461,17 @@ export default class ReactGridLayout extends React.Component<Props, State> {
...
+      l.w = w;
+      l.h = h;
 
       return l;
     });
 
     // Shouldn't ever happen, but typechecking makes it necessary
     if (!l) return;
 
-    let finalLayout;
```
dseif, line 439: I moved the majority of this logic to within `withLayoutItem`. This probably should have been within there to begin with, so sorry about missing that originally.

lib/ReactGridLayout.jsx
```diff
@@ -429,31 +461,17 @@ export default class ReactGridLayout extends React.Component<Props, State> {
...
-      if (["ne", "n", "nw"].indexOf(handle) !== -1) {
-        y = l.y + (l.h - h);
-        h = l.y !== y && y < 0 ? l.h : h;
-        y = y < 0 ? 0 : y;
-      }
-
-      l.w = w;
-      l.h = h;
+    finalLayout = newLayout;
+    if (shouldMoveItem) {
```
dseif, line 474: Similarily, we can't move the item within `withLayoutItem` which is why this is done outside of that block since we need the resulting `Layout` to perform the operation.

lib/ReactGridLayout.jsx
```diff
@@ -405,18 +405,50 @@ export default class ReactGridLayout extends React.Component<Props, State> {
...
+        }).filter(layoutItem => layoutItem.i !== l.i);
         hasCollisions = collisions.length > 0;
 
         // If we're colliding, we need adjust the placeholder.
         if (hasCollisions) {
+          // Reset layoutItem dimensions if there were collisions
+          y = l.y;
+          h = l.h;
+          x = l.x;
+          w = l.w;
```
dseif, line 451: This is the core of the change. If we detect that there was a collision *and* `preventCollision` is enabled, we reset the w/h/x/y values of the LayoutItem.

dseif:
To pass CI linter task, this will depend on #1938 being merged. Opted to fix those issues there instead of muddying this PR with that noise.

dseif:
@STRML if you have time this week would be great to get your eyes on this and #1938

STRML:
Thanks, I'll review this shortly. Please rebase on master.

lib/ReactGridLayout.jsx
```diff
@@ -429,31 +461,17 @@ export default class ReactGridLayout extends React.Component<Props, State> {
           if (Number.isFinite(leastY)) l.h = leastY - l.y;
         }
       }
+      l.w = w;
```
STRML, line 464: Doesn't this make L452-L461 essentially dead code?

test/spec/lifecycle-test.js
```diff
@@ -881,6 +881,24 @@ describe("Lifecycle tests", function () {
           });
         });
       });
+
+      it("Does not cause elements to move when resizing with preventCollision=true and no compaction (#1933)", () => {
```
STRML, line 885: Worth having a test where you can resize it (there's some empty space), as well as where you can resize it 1 grid unit but, say, not 2, because an item is in the way.

STRML (changes requested):
Minor comment & question

lib/ReactGridLayout.jsx
dseif: Yes, thanks for catching. 452-461 _shouldn't_ be needed anymore after testing, looks like the boundaries are still being calculated correctly.

test/spec/lifecycle-test.js
dseif: Sounds good, happy to add those cases as well.

test/spec/lifecycle-test.js
```diff
@@ -881,6 +881,100 @@ describe("Lifecycle tests", function () {
...
+          const gridLayout = wrapper.find("ReactGridLayout");
+          const itemId = 0;
+          const gridItem0 = findGridItemByText(gridLayout, itemId);
+          const handleElement = findHandleForGridItem(gridItem0, "e");
+          const pos = getCurrentPosition(gridItem0);
+          const positionBeforeResize = getGridItemData(gridLayout, itemId);
+
+          // Resize right two columns
+          resizeTo(handleElement, true, pos, pos.left + colWidth, pos.top);
+          resizeTo(handleElement, false, pos, pos.left + colWidth * 2, pos.top);
```
dseif, line 943: This duplication of `resizeTo` calls here is intentional. We first move 1 column over (which is valid, no collisions), then we move another column over (which is invalid, there is collisions). This is to better simulate an actual drag.

The problem was with the collision detection and the logic that's in place is that if we went immediately from currentX/currentY -> newX/newY without any valid positions inbetween, the "last known good X/Y are the original position. This isn't realistic behaviour when a user actually drags their mouse, which is why I simulated it like this.

A better solution is probably to look into using `DragEvent` instead of `MouseEvent` for these simulations. I can take a look into that outside of this PR to have tests better simulate reality.

STRML (approved):

lib/ReactGridLayout.jsx
```diff
@@ -405,55 +405,64 @@ export default class ReactGridLayout extends React.Component<Props, State> {
...
-            if (layoutItem.y > l.y) leastY = Math.min(leastY, layoutItem.y);
-          });
-
-          if (Number.isFinite(leastX)) l.w = leastX - l.x;
-          if (Number.isFinite(leastY)) l.h = leastY - l.y;
+          // Reset layoutItem dimensions if there were collisions
+          y = l.y;
+          h = l.h;
+          x = l.x;
+          w = l.w;
```
STRML, line 451:
```suggestion
          w = l.w;
          shouldMoveItem = false;
```

Should this not be done to short-circuit the move, since we know it's not happening?

dseif: Yea good suggestion, updated.

STRML:
Published in `1.4.2`

PR #2059, opened 2024-07-09
Fix calcXY calculation

### LINKED ISSUE #918: GridItem calcXY calculation is wrong

MaxMusing:
## Description

The current `calcXY` function results in items being placed further above and to the left of where they should be. This is more apparent when the margin is larger.

This PR fixes the calculation so that items are placed in the correct location.

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Related Tickets & Documents

Fixes #918

### Before

Notice how the ghost item is always to the left and above the hovered position.

https://github.com/react-grid-layout/react-grid-layout/assets/15393239/a9d3a27b-fafb-4322-bea1-3aeaaf184199

### After

Notice how the ghost item always snaps to the nearest available position.

https://github.com/react-grid-layout/react-grid-layout/assets/15393239/b6ccaaf0-aaea-4f57-aba4-16e814fd21f3

## Added tests?

- [x] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 examples

### DIFF
--- a/test/spec/lifecycle-test.js
+++ b/test/spec/lifecycle-test.js
@@ -375,14 +375,14 @@ describe("Lifecycle tests", function () {
         expect(gridLayout.state("droppingDOMNode")).toEqual(null);
 
         // Drag the droppable over the grid layout.
-        dragDroppableTo(wrapper, 200, 150);
+        dragDroppableTo(wrapper, 200, 140);
 
         // We should have the position in our state.
         expect(gridLayout.state("droppingPosition")).toHaveProperty(
           "left",
           200
         );
-        expect(gridLayout.state("droppingPosition")).toHaveProperty("top", 150);
+        expect(gridLayout.state("droppingPosition")).toHaveProperty("top", 140);
         // We should now have the placeholder element in our state.
         expect(gridLayout.state("droppingDOMNode")).toHaveProperty(
           "type",
--- a/lib/GridItem.jsx
+++ b/lib/GridItem.jsx
@@ -519,13 +519,7 @@ export default class GridItem extends React.Component<Props, State> {
     this.setState({ dragging: newPosition });
 
     // Call callback with this data
-    const { x, y } = calcXY(
-      positionParams,
-      top,
-      left,
-      w,
-      h
-    );
+    const { x, y } = calcXY(positionParams, top, left, w, h);
     return onDrag.call(this, i, x, y, {
       e,
       node,
@@ -550,13 +544,7 @@ export default class GridItem extends React.Component<Props, State> {
     const newPosition: PartialPosition = { top, left };
     this.setState({ dragging: null });
 
-    const { x, y } = calcXY(
-      this.getPositionParams(),
-      top,
-      left,
-      w,
-      h
-    );
+    const { x, y } = calcXY(this.getPositionParams(), top, left, w, h);
 
     return onDragStop.call(this, i, x, y, {
       e,
--- a/lib/GridItem.jsx
+++ b/lib/GridItem.jsx
@@ -519,11 +519,10 @@ export default class GridItem extends React.Component<Props, State> {
     this.setState({ dragging: newPosition });
 
     // Call callback with this data
-    const { containerPadding } = this.props;
     const { x, y } = calcXY(
       positionParams,
-      top - containerPadding[1],
-      left - containerPadding[0],
+      top,
+      left,
       w,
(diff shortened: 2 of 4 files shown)


test/spec/lifecycle-test.js
MaxMusing, line 1: This test now fails on `master` but passes on this branch.

STRML:
Thanks for submitting this.

After investigating, I believe both results are actually wrong, the old code and the new.

Assume a grid with `colWidth=100` and `margin=10`.

The layout will be:

```
-------------------------------------------
|  |         |  |         |  |         |  |
|10|   100   |10|   100   |10|   100   |10|
|  |         |  |         |  |         |  |
-------------------------------------------
```

So the midpoint of `x=0` is 60, `x=1` is 170, and so on.

The current math correctly calculates this:

```
let x = Math.round((left - margin[0]) / (colWidth + margin[0]));

// Assume `left = 60`, `x` should equal 0.5
Math.round((60 - 10) / (110));
Math.round(50 / 110); // 0.45

// Assume `left = 170`, `x` should equal 1.5
Math.round((170 - 10) / (110));
160 / 110 = 1.45
```

But the new math:

```js
let x = Math.round(left / (colWidth + margin[0]));
// Assume `left = 60`, `x` should equal 0.5
let x = Math.round(60 / (100 + 10));

// Assume `left = 170`, `x` should equal 1.5
Math.round(170 / (100 + 10));
170 / 110 = 1.54
```

The correct math would be:

```js
let x = Math.round((left - margin[0] / 2) / (colWidth + margin[0]));

// left = 60
Math.round(60 - 10 / 2) / (100 + 10);
Math.round(55 / 110) // 0.5

// left = 170
Math.round(170 - 10 / 2) / (100 + 10);
Math.round(165 / 110); // 1.5
```

MaxMusing:
Thanks for the reply! I went through the same thought process. I don't think the calculation is supposed to include margins at the start/end, since the `containerPadding` controls that (though it is set to `margin` by default).

For example, with `colWidth=100`, `margin=10`, `containerPadding=20`:

```
-------------------------------------------
|  |         |  |         |  |         |  |
|20|   100   |10|   100   |10|   100   |20|
|  |         |  |         |  |         |  |
-------------------------------------------
```

Here the midpoint of `x=0` is 70, `x=1` is 180, etc.

But, if you look at where the `calcXY` function is being called (e.g. `lib/GridItem.jsx` line 523), it already takes into account the container padding:

```js
const { x, y } = calcXY(
	positionParams,
	top - containerPadding[1],
	left - containerPadding[0],
	w,
	h
);
```

So, I believe the logic in the PR is correct. If you think it would be more clear, I can update `calcXY` to take `containerPadding` into account rather than relying on the caller to subtract it.

STRML:
That explains it, thanks! Yes I don't think the caller should be subtracting the padding, but this won't work if the container padding is different than the margin - it would need to be passed separately.

MaxMusing:
The latest commit subtracts the padding in the function rather than from the caller. I'm also going to quickly update the unit tests for `calcXY`.

MaxMusing:
Should be good to go, let me know if there's anything else you want changed @STRML.

STRML:
Small prettier issue, then it's good to merge

MaxMusing:
Should be fixed in the latest commit!

STRML:
Thanks for the very nice PR.

PR #2115, opened 2025-01-15
Memoize bottom computation in compact

JefHellemansBlockbax:
## Description

Hi, thank you for the amazing library. We were having some performance issues with grids that have a lot of items (we started seeing really laggy behaviour at around 1200). A combination of a couple of fixes took care of almost all of our problems, namely:
1. The `requestAnimationFrame` fix in `react-draggable`. In our use case, we have clones of grid items which get removed when dragging starts for visual clarity. I've extended upon that with another PR https://github.com/react-grid-layout/react-draggable/pull/773. It would be great if `react-grid-layout` can eventually migrate to a newer `react-draggable` version.
1. The `flushSync` fix in this PR https://github.com/react-grid-layout/react-grid-layout/pull/2043.
1. The low-hanging fruit performance optimization that I'm suggesting here.

While 1 and 2 got us very far, this small memoization helps reduce the overhead of the `compact` function, which is executed every `onDrag` so tends to get expensive at this scale.

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [x] 🔥 Performance Improvements
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 examples

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -239,6 +239,8 @@ export function compact(
 ): Layout {
   // Statics go in the compareWith array right away so items flow around them.
   const compareWith = getStatics(layout);
+  // We keep track of the bottom position.
+  let b = bottom(compareWith);
   // We go through the items by row and column.
   const sorted = sortLayoutItems(layout, compactType);
   // Holding for new items.
@@ -249,7 +251,8 @@ export function compact(
 
     // Don't move static elements
     if (!l.static) {
-      l = compactItem(compareWith, l, compactType, cols, sorted, allowOverlap);
+      l = compactItem(compareWith, l, compactType, cols, sorted, allowOverlap, b);
+      b = Math.max(b, l.y + l.h);
 
       // Add to comparison array. We only collide with items before this one.
       // Statics are already in this array.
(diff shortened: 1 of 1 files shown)


STRML:
This is nice. I will merge it for now, but it may be superseded by #2152.

JefHellemansBlockbax:
Hi @STRML, thank you for accepting this PR. I noticed it's not part of the current release, and that while #2152 has moved forward, it's not quite ready yet. Was there a particular reason this was not included in any of the releases?

Looking at this PR again, I do spot a flaw with the API of `compactItem` after this change though. `allowOverlap` is an optional parameter, followed by the new `b` param which is mandatory. We can probably use the old logic as a fallback if we want to patch this.

STRML:
Ah yes. If you'd propose a fix that'd be helpful. Thank you @JefHellemansBlockbax.

PR #2150, opened 2025-05-19
fix(utils): fix case when margins are not consistent due to rounding

akmed0zmey:
## Description

Currently margins are not consistent. For example if we set margins to `[1,1]` it sometimes makes it 2px or 0px (see screenshots). This PR introduces margins check and fix.

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

### Before fix
Screenshot from 2025-05-19 15-56-41
Screenshot from 2025-05-19 15-54-42

### After fix
Screenshot from 2025-05-19 15-58-22

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 examples

### DIFF
--- a/lib/calculateUtils.js
+++ b/lib/calculateUtils.js
@@ -86,6 +86,34 @@ export function calcGridItemPosition(
     out.left = Math.round((colWidth + margin[0]) * x + containerPadding[0]);
   }
 
+  if (!state || !(state.dragging || state.resizing)) {
+    if (Number.isFinite(w)) {
+      const marginRight =
+        Math.round((colWidth + margin[0]) * (x + w) + containerPadding[0]) -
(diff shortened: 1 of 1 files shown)


STRML:
Hi, could you add a test for this that fails under the old code?

akmed0zmey:
> Hi, could you add a test for this that fails under the old code?

Done!

PR #2152, opened 2025-06-16
perf(compact): implement faster vertical compaction

morris:
## Description

This PR implements fast vertical compaction using a "rising tide" algorithm. For all practical cases, the algorithm has linear time complexity, and requires almost zero additional memory allocations (after cloning and sorting the layout).

While working with larger layouts (200+ items), dragging and dropping items felt sluggish, and I found `compact()` to be one of the major bottlenecks with lots of recursion and allocations/garbage collection. The following two screenshots from the Chrome profiler should give an idea of the improvement.

Current `master`:

Screenshot 2025-06-16 at 18 36 51

New algorithm:

Screenshot 2025-06-16 at 18 39 15

Note that the comparison is not fully representative. The recordings were done on random, distinct 300-item layouts (one shot).

Let me know if there's interest, happy to add explanations and/or tests :)

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [x] 🔥 Performance Improvements
- [x] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Added tests?

- [x] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 examples

### DIFF
--- a/test/spec/utils-test.js
+++ b/test/spec/utils-test.js
@@ -459,7 +459,7 @@ describe("compact vertical", () => {
     });
   });
 
-  it("Does not produce overlaps for random layouts", () => {
+  it("Does not produce overlaps for random layouts, and is idemptotent", () => {
     // Increase this during dev to test more layouts.
     const numLayouts = 10;
 
@@ -502,6 +502,11 @@ describe("compact vertical", () => {
                 }
               }
             }
+
+            // Test for idempotence
+            const out2 = compact(out, "vertical", cols);
+
+            expect(out2).toEqual(out);
           }
         }
       }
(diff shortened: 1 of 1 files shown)


STRML:
Hi,

This is very interesting! Thanks for implementing.

This changes the behavior of the default compactor, does it not? Or are the results the same and this is purely performance?

STRML:
Additionally, can you see this being implemented horizontally too, and if so, can we remove a bunch of old code? I would be very happy with that.

morris:
Hi!

> This changes the behavior of the default compactor

I think it does, but hopefully only in subtle ways (weird edge cases). I'll try to quantify this with some tests, let me get back to you.

> can you see this being implemented horizontally too

I haven't tried, but I suspect horizontal compaction needs a different approach. While grids can grow vertically indefinitely, the number of columns is constant, so it looks like a (slightly?) separate problem. I'm thinking t might be safer to take it on separately 🤔 would you see this as a show stopper for this PR?

STRML:
I don't want to leave the code base with a lot of dead code, so it's fairly high priority. I think this is a measurable improvement and the change is actually easier to read, so I'm excited about what it might bring.

morris:
I've added two minimal tests illustrating differences in behavior. The first one produces a more compact layout, the second one produces a less compact result (which may be considered a regression) than current `master`. I think the input layouts are a bit ambiguous though, so not sure if either behavior is "better" or "worse".

Running some randomized tests (roughly same inputs from the new fuzzy tests), I've found the new compaction algorithm to produce equivalent layouts for about 65-75% of inputs. This may be indeed too much of a deviation.

However, I think it comes down more to the user experience while moving items: Does the layout change in unexpected/unintended ways while dragging and dropping? Regarding that, I'm fairly confident it works well (if not better):

https://github.com/user-attachments/assets/11b3abcb-ae03-4a42-a4df-cbfa4bf4d7a7

In the end, it's your call of course :) let me know if you think the changed behavior is good enough. I might take a stab at horizontal compaction in that case

morris:
Playing with it more, I found the algorithm to be surprisingly powerful as it can support manipulation (move/resize) as well by applying the manipulation, sorting cleverly, and using compaction as a repair step.

I was able to implement a RGL alternative that's more opinionated (less options, e.g. no horizontal compaction) but much, much faster, based on it: https://github.com/morris/fast-grid-layout - may be interesting for you as well :)

STRML:
That's pretty cool @morris. I'll spend some time with this again.

I'm a performance junkie and you really whipped my butt with this one.

STRML:
## Auto code review

No issues found. Checked for bugs and CLAUDE.md compliance.

**Note:** This PR was created before the v2 TypeScript rewrite and targets the legacy `lib/utils.js` codebase. The master branch has since been updated to v2 with TypeScript in `src/core/`. To merge this performance improvement, the changes would need to be rebased and reimplemented in `src/core/compactors.ts`.

STRML:
Ok this is pretty good. 

```
┌───────────────────────────────────────────────────────────────────────────┐
│                    ⚡ Compactor Performance Comparison                     │
├───────────────────────────────────────────────────────────────────────────┤
│ Items    │ Standard Vertical     │ Fast Vertical         │ Speedup        │
├──────────┼───────────────────────┼───────────────────────┼────────────────┤
│ 50       │             105.03 µs │              17.77 µs │   5.91x faster │
│ 100      │             261.62 µs │              36.20 µs │   7.23x faster │
│ 200      │             928.55 µs │              55.97 µs │  16.59x faster │
│ 500      │               6.56 ms │             141.10 µs │  46.46x faster │
└───────────────────────────────────────────────────────────────────────────┘
```

STRML:
This algorithm has been migrated to the v2 TypeScript architecture in #2184. Thanks for the contribution @morris.

morris:
Awesome, thanks for incorporating it 😌

PR #2196, opened 2025-12-14
fix: center dropped items on cursor position

STRML:
## Summary
- Centers dropped items on the cursor position instead of treating cursor as top-left corner
- Adds `dragOffsetX` support to `onDropDragOver` callback for cursor offset awareness
- Clamps to grid bounds to prevent negative positioning

## Problem
When dragging items from outside the grid, the placeholder appears offset from the cursor, creating a misaligned, unresponsive experience. The library treats the cursor as the top-left corner of the dragged item, making the gap between cursor and placeholder especially noticeable with larger items.

## Solution
Implement item centering by offsetting the dropping position by half the item's dimensions. This centers the placeholder on the cursor, providing a more natural drag-and-drop experience.

The `onDropDragOver` callback can now also return a `dragOffsetX` value for cursor offset awareness - this preserves the exact position where users grab the item, mirroring how native drag-and-drop works.

This is a port of PR #2167 to the v2 codebase.

## Test plan
- [x] Added unit tests for centering calculation
- [x] All existing tests pass
- [x] Manual testing with drag-from-outside example

### DIFF
--- a/src/react/components/GridLayout.tsx
+++ b/src/react/components/GridLayout.tsx
@@ -855,8 +855,8 @@ export function GridLayout(props: GridLayoutProps): ReactElement {
         ]);
       } else if (droppingPosition) {
         const shouldUpdate =
-          droppingPosition.left !== clampedGridX ||
-          droppingPosition.top !== clampedGridY;
+          droppingPosition.left !== newDroppingPosition.left ||
+          droppingPosition.top !== newDroppingPosition.top;
         if (shouldUpdate) {
           setDroppingPosition(newDroppingPosition);
         }


STRML:
## Code review

Found 1 issue:

1. Incorrect scale comparison in `shouldUpdate` check (bug): The `shouldUpdate` comparison compares `droppingPosition.left/top` (which are stored as `clampedGridX / transformScale`) against raw `clampedGridX/clampedGridY` values. When `transformScale !== 1`, this will always evaluate to `true`, causing unnecessary state updates and re-renders.

https://github.com/react-grid-layout/react-grid-layout/blob/e702bdaffce43690db8a09760af478e03b8263c3/src/react/components/GridLayout.tsx#L854-L860

**Fix:** Compare against scaled values:
```typescript
const shouldUpdate =
  droppingPosition.left !== newDroppingPosition.left ||
  droppingPosition.top !== newDroppingPosition.top;
```

PR #2216, opened 2025-12-29
fix: use custom compactor.compact() method when provided (#2213)

### LINKED ISSUE #2213: Can't use custom compactors
### Describe the bug

https://github.com/react-grid-layout/react-grid-layout?tab=readme-ov-file#extending-custom-compactors--position-strategies

I'm trying to use a custom compactor as described in the docs here. 
```tsx
const customCompactor = {
  type: "custom",
  allowOverlap: false,
  preventCollision: false,
  compact(layout, cols) {
    console.log("compact", layout, cols);
    return layout;
  },
  onMove(layout, item, x, y, cols) {
    console.log("onMove", layout, item, x, y, cols);
    return layout;
  }
};

<ResponsiveGridLayout
   compactor={customCompactor}
   ...
   />
```

`allowOverlap` and `preventCollision` are respected but compact or onMove functions are never executed.

https://codesandbox.io/p/sandbox/old-fog-pm8zp8

1. Create a custom compactor with Compactor interface.
2. Pass it to `ReactGridLayout`
3. Move around tiles or resize the window. Custom compact function is never called.

### Expected behavior

I expect the custom compact function to be used.

2.1.1

### Operating System Version

macOS

### Browser

Chrome

### Additional context

_No response_

### Screenshots or Videos

_No response_

STRML:
## Summary

Fixes #2213

The `Compactor` interface defines `compact()` and `onMove()` methods that allow users to implement custom compaction algorithms. However, `GridLayout` was ignoring these methods and only using the legacy `compact()` function from `compact-compat.js`. This meant custom compactors like the one in the issue would never have their `compact()` method called.

### Root cause

`GridLayout.tsx` was:
1. Extracting only properties from the compactor (`type`, `allowOverlap`, `preventCollision`)
2. Passing these to the legacy `compact()` function
3. Never calling `compactor.compact()` or `compactor.onMove()`

### The fix

Replace all calls to the legacy `compact()` function with calls to `compactor.compact()`:
- `synchronizeLayoutWithChildren` now takes a `Compactor` instead of `compactType`/`allowOverlap`
- All `setLayout` calls that used `compact()` now use `compactor.compact()`
- Updated `useCallback` dependency arrays accordingly

The built-in compactors (like `verticalCompactor`, `horizontalCompactor`) already implement `allowOverlap` behavior internally via their `compact()` method, so the allowOverlap check before calling compact can be removed - the compactor handles it.

## Test plan

- [x] Added test that fails without fix (custom `compact()` never called)
- [x] Test passes with fix
- [x] All existing tests pass (519 passed)

### DIFF
--- a/src/react/components/GridLayout.tsx
+++ b/src/react/components/GridLayout.tsx
@@ -49,7 +49,7 @@ import {
   correctBounds
 } from "../../core/layout.js";
 import { getAllCollisions } from "../../core/collision.js";
-import { compact } from "../../core/compact-compat.js";
+// Note: compact from compact-compat.js is NOT used - we use compactor.compact() instead (#2213)
 import { getCompactor } from "../../core/compactors.js";
 import {
   calcXY,
@@ -239,8 +239,7 @@ function synchronizeLayoutWithChildren(
   initialLayout: Layout,
   children: React.ReactNode,
   cols: number,
-  compactType: CompactType,
-  allowOverlap: boolean
+  compactor: Compactor
 ): Layout {
   const layout: LayoutItem[] = [];
   const childKeys = new Set<string>();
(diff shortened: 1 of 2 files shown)


STRML:
## Auto code review

No issues found. Checked for bugs and CLAUDE.md compliance.

Issue #2221, opened 2025-12-30
The display position of the dragged element is abnormal


Pzx1997:
### Describe the bug

I'm considering whether to use this library. I found this issue when looking at the official demo. Is this correct?

https://github.com/user-attachments/assets/e21bfe85-424f-4dcc-8234-39bea5988ba1

### Your Example Website or App

https://github.com/user-attachments/assets/e21bfe85-424f-4dcc-8234-39bea5988ba1

### Steps to Reproduce the Bug or Issue

All cases on the official website can be reproduced by clicking on the elements.

### Expected behavior

The position of the dragged element should follow the mouse.

### react-grid-layout library version

2.2.0

### Operating System Version

macOS

### Browser

Chrome

STRML:
This is fixed in v2.2.0! The "placeholder appearing on click" behavior was addressed with the new `dragConfig.threshold` option.

In v2, the default threshold is 3px - the drag placeholder only appears after the mouse has moved beyond this threshold:

```jsx
<GridLayout dragConfig={{ threshold: 3 }} ... />
```

Please test with v2.2.0+ where clicking without dragging no longer shows the placeholder.

pawojciechowski:
@STRML it's not fixed - I'm experiencing the same issue. The problem is that when the main app container is relative & positioned not at x: 0, y: 0 (e.g. you have a sidebar), then after the click the drag placeholder is wrongly positioned via transform (it seems the transform is used based on the browser window, not the actual container position). Can reproduce using 2.2.0 with threshold 3 & 10.

pawojciechowski:
The problem is with this fragment:

https://github.com/react-grid-layout/react-grid-layout/blob/9ab22eaf3e8c9faf62113cfcb0cae327ecf6d8ca/src/react/components/GridItem.tsx#L381-L388

& the default strategy:

https://github.com/react-grid-layout/react-grid-layout/blob/9ab22eaf3e8c9faf62113cfcb0cae327ecf6d8ca/src/core/position.ts#L275-L286

It does not adjust for parent at all.

pawojciechowski:
Thanks for the fix!

STRML:
Fixed, releasing 2.2.1 now

PR #2223, opened 2025-12-30
fix: remove calcDragPosition from default strategies to fix drag position jump

STRML:
## Summary

- Fixes a bug where items would jump half a screen down when drag started
- Removes `calcDragPosition` from `transformStrategy` and `absoluteStrategy`
- Makes `calcDragPosition` optional in the `PositionStrategy` interface
- Includes regression test that verifies the fix

## Root Cause

The bug was introduced in commit 0c80a9f which added the `PositionStrategy` extension point. The default `transformStrategy` and `absoluteStrategy` had `calcDragPosition` methods that returned screen coordinates (`clientX - offsetX, clientY - offsetY`), but the drag system expects parent-relative coordinates.

When the grid container was offset from the top of the page (e.g., 500px down), clicking on an item would incorrectly position it at screen coordinates (y=500+) instead of parent-relative (y=0-10), causing the item to jump.

## Fix

Remove `calcDragPosition` from the default strategies, allowing react-draggable to use its built-in parent-relative coordinate calculation. Only `createScaledStrategy` retains `calcDragPosition` since it needs to adjust for the scale factor.

## Test plan

- [x] Added regression test that mocks `getBoundingClientRect` to simulate grid offset from page top
- [x] Verified test fails with buggy code (`y=13` instead of `< 5`)
- [x] Verified test passes with fix
- [x] All 527 tests pass

### DIFF
--- a/src/core/position.ts
+++ b/src/core/position.ts
@@ -270,18 +270,6 @@ export const transformStrategy: PositionStrategy = {
 
   calcStyle(pos: Position): React.CSSProperties {
     return setTransform(pos) as React.CSSProperties;
-  },
-
-  calcDragPosition(
-    clientX: number,
-    clientY: number,
-    offsetX: number,
-    offsetY: number
-  ): PartialPosition {
-    return {
-      left: clientX - offsetX,
-      top: clientY - offsetY
-    };
   }
 };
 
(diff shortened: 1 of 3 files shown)


STRML:
Fixes #2221

Issue #2109, opened 2024-12-30
A quick uncommitted item resize would result into a collapsed item size by width or height


dmitrylen-fieldpulse:
### Describe the bug

Encountered a buggy behaviour that makes UI looking unexpected – when quickly resizing an item by e.g. **NW, W, or SW** without releasing the mouse button to commit the change, the resizeable item would experience its size collapsing by width or height.

### Your Example Website or App

https://react-grid-layout.github.io/react-grid-layout/examples/20-resizable-handles.html

### Steps to Reproduce the Bug or Issue

1. Use sandbox from the documentation (https://react-grid-layout.github.io/react-grid-layout/examples/20-resizable-handles.html)
2. Mouse down on a resize handler
3.  Quick move to the left without release a mouse button to commit the resizing result
AR: The resizeable item size is collapsed by width or height during a quick uncommitted resizing

### Expected behavior

The resizeable item size should be preserved during the resizing

### react-grid-layout library version

v1.5.0 (reproducible on ^1.3.5 as well)

### Operating System Version

macOS, Windows (should be an issue as well)

### Browser

Chrome

### Screenshots or Videos

https://github.com/user-attachments/assets/0c5da003-7b2e-4c26-951a-69b934c47673

dmitrylen-fieldpulse:
Ping

piechoo:
I'm also facing this issue, its easily reproductible with quickly resizing far-right element to the left, any updates ?

MaxLascombe:
Same here. Was facing the same issue, but we decided to simply ignore it for now. It would be nice for this to be fixed tho.

Has either of you fixed this in a fork? Would there be value in me trying to fix it and sharing the fix?

dmitrylen-fieldpulse:
@MaxLascombe , No, we haven't fixed this issue. Specifically for our use case, the issue has "Lowest" priority, so there is not much value for us 🥹 

If you do come up with a solution, please share it here – it could still be helpful for maintainers and others facing the same issue ⭐ 

@piechoo , please check, is the solution would be actual to you?

TheWeirdDev:
We're facing the same issue.

NJCREID:
I've spent some time debugging because this was bothering me and i've manage to fix it.:

### lib/GridItem.tsx, render(): 
In the render function of GridItem.tsx mixinResizable is called (pos being the initial position passed by the user at this point):
```typescript
  newChild = this.mixinResizable(newChild, pos, isResizable);
[...]

PR #2224, opened 2025-12-30
fix: west resize not persisting final size due to React batching

STRML:
## Summary
- Fix `onResizeStop` not persisting the final resize size due to React state batching
- Store resize position during `onResize` and use it in `onResizeStop` instead of stale data from react-resizable
- Re-apply resize logic in `onResizeStop` handler to ensure final layout is correct

## Problem
When resizing from west handles (W, NW, SW), the resize would appear to work during the drag, but could snap back to incorrect values on mouse release. This was caused by:

1. `react-resizable`'s `onResizeStop` reports the DOM element's current size, but React hasn't re-rendered yet (state updates are batched), so it reports stale size data
2. `GridLayout.onResizeStop` was reading from `layoutRef.current` which still contained the pre-resize layout

## Solution
- **GridItem**: Store resize position during `onResize` in a ref, use it in `onResizeStop` instead of `data.size`
- **GridLayout**: Re-apply the same resize logic in `onResizeStop` as `onResize` (including collision checking for `preventCollision` mode)

## Test plan
- [x] Added regression test for west resize with item at y=1
- [x] All existing tests pass
- [x] `preventCollision` tests still pass

Note: This is related to but does not fully address #2109. The original issue describes position shifting *during* resize, which may be a separate bug to investigate.

### DIFF
--- a/.claude/agents/code-reviewer.md
+++ b/.claude/agents/code-reviewer.md
@@ -1,0 +1,33 @@
+---
+name: code-reviewer
+description: Expert code review specialist for quality, security, and maintainability. Use PROACTIVELY after writing or modifying code to ensure high development standards.
+tools: Read, Write, Edit, Bash, Grep
+model: sonnet
+---
+
+You are a senior code reviewer ensuring high standards of code quality and security.
+
+When invoked:
+
+1. Run git diff to see recent changes
+2. Focus on modified files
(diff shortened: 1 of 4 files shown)


STRML:
This is fixed by react-resizable@3.1.0.

PR #2245, opened 2026-02-27
fix: prevent resize drift caused by React render latency

WhiteMinds:
## Description

This PR fixes progressive resize drift where widgets resize slower than mouse movement. The further you drag a resize handle, the bigger the offset becomes.

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## Root Cause

When the user drags a resize handle, \`react-resizable\` computes the new size as:

```js
width = this.props.width + deltaX
```

\`this.props.width\` comes from the last React render. When React can't re-render between consecutive \`mousemove\` events (which is common during fast dragging), multiple events share the same stale \`props.width\`:

```
Render #1: props.width = 200
  mousemove A: delta=3 → size = 200+3 = 203 → stored in resizePositionRef
  mousemove B: delta=2 → size = 200+2 = 202 → overwrites resizePositionRef (203 is lost!)
Render #2: props.width = 202 (lost 3px of delta from event A)
```

Each skipped render loses accumulated deltas, causing the widget to progressively lag behind the mouse cursor.

## Fix

Track the \`width\`/\`height\` that was last rendered to \`<Resizable>\` as props via a ref (\`lastRenderedResizeSizeRef\`). In the resize handler, instead of using Resizable's absolute \`size\` directly, extract the true delta relative to the last rendered value and accumulate it onto \`resizePositionRef\`:

```ts
const delta = size.width - lastRenderedResizeSizeRef.current.width;
correctedWidth = resizePositionRef.current.width + delta;
```

This way, even if React skips renders between mouse events, no delta is lost — each event's contribution is correctly accumulated.

The fix is contained entirely within \`GridItem.tsx\` (~30 lines) and does not modify \`react-resizable\` or \`react-draggable\`.

## Mobile & Desktop Screenshots/Recordings

N/A — no visual changes beyond the bug fix (resize now tracks the mouse cursor accurately).

## Added tests?

- [ ] 👍 yes

The bug depends on React render timing relative to DOM mousemove events, which cannot be reliably reproduced in unit tests. All 517 existing tests pass.

## Added to documentation?

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved accuracy of grid item resizing calculations to ensure correct size adjustments persist even when rendering updates occur between consecutive mouse events.

### DIFF
--- a/src/react/components/GridItem.tsx
+++ b/src/react/components/GridItem.tsx
@@ -249,6 +249,13 @@ export function GridItem(props: GridItemProps): ReactElement {
     width: 0,
     height: 0
   });
+  // Track what width/height was last rendered to Resizable as props.
+  // This is needed to correctly extract deltas when React skips renders
+  // between mouse events (Resizable uses stale props.width + deltaX).
+  const lastRenderedResizeSizeRef = useRef<{
+    width: number;
+    height: number;
+  }>({ width: 0, height: 0 });
 
   // Previous dropping position for comparison
   const prevDroppingPositionRef = useRef<DroppingPosition | undefined>(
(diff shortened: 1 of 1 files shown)


WhiteMinds:
Closing this PR. The root cause is in react-resizable, not react-grid-layout. Filed the fix upstream: https://github.com/react-grid-layout/react-resizable/pull/255

Issue #2281, opened 2026-08-05
Meta: implement touch external-drop adapter (#1756)


STRML:
## Meta issue for implementing #1756

External HTML5 drag-and-drop (drop-from-outside) doesn't work on touch/mobile because touch events never produce dragover/drop. This is the one ready-for-agent issue not shipped in the 2026-08 sweep.

### Design sketch

A touch-to-drop adapter that translates touch events into the existing drop pipeline. The grid's drop math (handleDragOver, calcXY, droppingItem placeholder) is already correct - only the event source is wrong.

Minimum viable:
1. Capture touchstart on a drag source element (or document-level touch listeners).
2. On touchmove, translate clientX/Y -> clampedGridX/Y -> calcXY -> render the droppingItem placeholder.
3. On touchend, commit via the existing onDrop callback.

### Open questions
- Should the touch adapter be part of GridLayout (inline onTouchMove handlers) or a separate component?
- How should the user designate a touch-draggable source? (New prop touchDragSource, or reuse HTML5 draggable=true?)
- Should this be a scoped feature behind a dropConfig flag or default-on?

### Prior work
The agent brief from triage lives on #1756. The edge-scroll (#2232) was split from this - touch-drop is a distinct input problem, not a layout problem.

### Testing
Needs real-device verification (iOS Safari, Android Chrome). A jest test can simulate touch events through the adapter; a Playwright e2e test (using the harness from #2275) would be the real win.

> *This was generated by AI during triage.*

STRML:
Shipped in #2282 (merged 2026-08-05). Touch external-drop now works out of the box on mobile.

To use it, add `data-rgl-draggable={true}` to the element you already drag. The grid listens for touchstart on that element, then translates touchmove/touchend into the drop pipeline: the placeholder tracks the finger, touchend over the grid commits via `onDrop`, and leaving the grid removes the placeholder. Enabled whenever `dropConfig.enabled` is on (`touchEnabled` defaults to true). Desktop HTML5 drag-and-drop is unchanged.

Docs and config:
- `dropConfig.touchEnabled` — default `true`
- `dropConfig.touchDragSource` — default `[data-rgl-draggable]`
- Unit tests: 10 (mock touch events)
- E2E specs included but marked `fixme` — headless Chromium doesn't fire DOM touch events from CDP `Input.dispatchTouchEvent`, so they need a headed run or real-device pass. That's the one gap worth confirming.

Closing both this and #1756.

Issue #1756, opened 2022-07-25
react-grid-layout "drag from outside" not working in mobile view


NityaJain1099:
### Describe the bug

Hello,

The react-grid-layout "drag from outside" is not working in mobile resolution. I have also attached the code sandbox link below for the reference.

I request to please look into the issue or please suggest if there is any solution to make it work in mobile view.

Thanks!

### Your Example Website or App

https://codesandbox.io/s/react-grid-layout-nested-grids-drag-from-outside-forked-hs06sm?file=/dragFromOutside.js

### Steps to Reproduce the Bug or Issue

1. Go to code sandbox link provided above.
2. Open the console.
3. Click on the "Toggle Device Toolbar" (for the mobile view).
4. The "drag drop from outside" is not working in mobile resolution.

### Expected behavior

The "drag from outside" example is working in normal laptop screens but its not working in mobile resolution.

### react-grid-layout library version

1.3.4

### Operating System Version

Windows

### Browser

Chrome

ljluestc:
```

import React, { useState } from "react";
import GridLayout from "react-grid-layout";
import "./styles.css";
import { WidthProvider } from "react-grid-layout";

const ResponsiveGridLayout = WidthProvider(GridLayout);

const dragFromOutsideItem = {
  w: 2,
  h: 2,
  x: 0,
  y: 0,
  i: "new-item"
};

export default function DragFromOutside() {
  const [layout, setLayout] = useState([]);

  const onDrop = (layout, item, e) => {
    setLayout([...layout, { ...dragFromOutsideItem, i: `new-item-${layout.length}` }]);
  };

  return (
    <div>
      <div
        className="draggable-item"
        style={{
          width: "100px",
          height: "100px",
          backgroundColor: "lightblue",
          margin: "10px",
          cursor: "pointer",
          touchAction: "none" // Enables touch events
        }}
        onTouchStart={(e) => {
          // Manually start drag on touch start
          e.dataTransfer = new DataTransfer();
          e.dataTransfer.setData("text/plain", "");
        }}
        draggable
      >
        Drag me
      </div>
      <ResponsiveGridLayout
        className="layout"
        layout={layout}
        cols={12}
        rowHeight={30}
        width={1200}
        isDroppable={true}
        onDrop={onDrop}
        measureBeforeMount={false}
        compactType={null}
        preventCollision={true}
      >
        {layout.map((item) => (
          <div key={item.i} className="grid-item">
            {item.i}
[...]

PR #2282, opened 2026-08-05
feat: touch external-drop adapter (#2281)

STRML:
## Summary

External drag-and-drop (drop-from-outside) never worked on mobile. iOS Safari and Android Chrome don't fire `dragover`/`drop` from a touch-dragged source, so the placeholder never appeared and `onDrop` never fired. This PR adds a touch adapter that translates touch events on a marked source element into the existing drop pipeline.

**To use it:** add `data-rgl-draggable` to your drag source element. That's it. The grid listens for `touchstart` on marked elements, then `touchmove`/`touchend` at the document level - a touch that starts on a sibling retargets `touchmove` to the element where `touchstart` fired, so the grid's own handlers would never see it. `touchmove` puts the dropping placeholder under the finger, `touchend` over the grid commits via `onDrop`, and leaving the grid removes the placeholder.

It's on whenever `dropConfig.enabled` is (`touchEnabled` defaults to `true`). Desktop is unchanged.

This was the last ready-for-agent item from the 2026-08 triage sweep (#2281, meta for #1756).

## Changes

- `DropConfig` gains `touchEnabled` (default `true`) and `touchDragSource` (default `[data-rgl-draggable]`)
- `dropMath.ts` (new): `computeDropPosition` extracted from `handleDragOver`. The drop geometry is now shared between desktop and touch instead of faking a React drag event.
- `GridLayout.tsx`: document-level touch listeners. `applyTouchDropPosition` reads placeholder state through refs so the listeners bind once.
- Tests: 10 unit (mock touch events through the adapter) + 2 Playwright e2e (CDP `Input.dispatchTouchEvent`)
- Example: `data-rgl-draggable` on the existing drag-from-outside source

## Status — 2026-08-05

| Step | State |
|---|---|
| Unit tests | 580 pass (570 + 10 new), tsc clean |
| E2E | Touch specs marked `fixme` (CDP touch events don't fire in headless Chromium); 4 smoke specs run |
| Formatting/lint | Prettier clean (byte-identical on all 7 changed files), ESLint 0 errors |
| CI | Linters ✅ · Audit ✅ · labeler ✅ · build re-running on 36117ed |

CodeRabbit findings addressed: named `GridLayout` import, `data-rgl-draggable={true}`, `touchDragSource` doc, prettier. Pushed back on `onDragOver`-during-touch (no native `DragEvent` on touch; scoped as a follow-up) and ref-mirror-in-render (matches existing `layoutRef` pattern).
## Open follow-ups

- Real-device pass (iOS Safari, Android Chrome) to confirm `elementFromPoint` containment + scroll behavior
- `dropConfig.onDragOver` isn't consulted during touch - it takes a native `DragEvent`, which touch never produces. Touch-specific size/offset control would be a follow-up.

## Test plan

- [x] 580 unit tests pass
- [x] `tsc --noEmit` clean
- [x] `make build` succeeds
- [x] Prettier + ESLint clean
- [x] E2E harness builds

## Summary by CodeRabbit

* **New Features**
  * Added touch-based external dragging and dropping for grid layouts.
  * Touch dragging displays a live placeholder and commits items when released over the grid.
  * Added settings to enable touch support and identify draggable source elements.
  * Touch dropping is enabled by default for marked draggable elements.

* **Bug Fixes**
  * Prevented canceled or incomplete touch gestures from creating dropped items.
  * Preserved existing desktop drag-and-drop behavior.

### DIFF
--- a/src/core/types.ts
+++ b/src/core/types.ts
@@ -576,7 +576,7 @@ export interface DropConfig {
    * CSS selector for elements that can be touch-dragged onto the grid.
    * The adapter only listens for touchstart on matching elements, so an
    * explicit marker prevents surprise scroll-hijacking.
-   * Add `data-rgl-draggable` (the default) to your drag source element.
+   * Defaults to `[data-rgl-draggable]` - add that attribute to your source.
    */
   touchDragSource?: string;
 }
--- a/src/core/types.ts
+++ b/src/core/types.ts
@@ -563,12 +563,30 @@ export interface DropConfig {
     | { w?: number; h?: number; dragOffsetX?: number; dragOffsetY?: number }
     | false
     | void;
+
+  /**
+   * Enable touch-to-drop for external drag sources on mobile.
+   * Touch events never fire HTML5 dragover/drop, so the grid translates
+   * touchmove/touchend into the drop pipeline when this is on.
+   * Only elements matching `touchDragSource` are captured (default: true).
+   */
+  touchEnabled?: boolean;
+
+  /**
+   * CSS selector for elements that can be touch-dragged onto the grid.
+   * The adapter only listens for touchstart on matching elements, so an
+   * explicit marker prevents surprise scroll-hijacking.
+   * Add `data-rgl-draggable` (the default) to your drag source element.
+   */
+  touchDragSource?: string;
 }
 
 /** Default drop configuration */
 export const defaultDropConfig: DropConfig = {
   enabled: false,
-  defaultItem: { w: 1, h: 1 }
+  defaultItem: { w: 1, h: 1 },
+  touchEnabled: true,
+  touchDragSource: "[data-rgl-draggable]"
 };
 
 // ============================================================================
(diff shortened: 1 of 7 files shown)


STRML:
> **Unresolved: onDragOver during touch**

Your point about `dropConfigOnDragOver`/`onDropDragOverProp` not being consulted during touch is correct - touch never produces a native `DragEvent`, so `onDragOver`'s contract (cursor-offset and dimension-override tweaking off a desktop DnD event) doesn't apply. The finger position determines placement directly through the same `computeDropPosition` math both paths share.

A touch-specific override API could be useful, but it should be scoped to what touch can express (position, not dataTransfer). That's listed as an open follow-up in the PR body; I'd rather not block the initial merge on a superset API.

Also addressing: the ref-mirror suggestion. I sync `droppingDOMNodeRef`/`droppingPositionRef` during render to match the existing `layoutRef` pattern in this component (`layoutRef.current = layout` at the same point). The touch handlers read through refs so document listeners bind once. I'm keeping render-phase sync for consistency rather than introducing `useLayoutEffect` for just these two.

STRML:
@coderabbitai review

PR #2284, opened 2026-08-05
refactor: dedup drop pipeline between desktop and touch

STRML:
## Summary

Follow-up refactor to the touch external-drop adapter (#2282): dedup the drop pipeline so the desktop HTML5 drag-over path and the touch adapter share the same placeholder-placement and drop-commit code.

## Changes

- Extract `placeDroppingItem()` — the create-or-update placeholder block was copy-pasted between `handleDragOver` and the touch adapter; now both call the same helper.
- Extract `commitDrop()` — `handleTouchEnd` replicated `handleDrop`'s commit verbatim and had already drifted (it skipped the `dragEnterCounterRef` reset). One shared commit function removes the divergence.
- `onDropProp` mirrored through a ref so the document touch listeners don't rebind when the consumer passes an inline `onDrop` (new identity every parent render).
- Stable `setGridRef` useCallback — the inline ref callback was re-invoked (null → node) on every render, which during a touch drag momentarily nulled the grid ref per move.
- `DropPositionInput.droppingItem` narrowed to `{ w, h }` (the `i` field was never read).
- `getTouch` hardcodes `changedTouches` instead of threading an always-same string param.

## Verification

- 580+ unit tests pass (touch-drop, drop-alignment, responsive, margin suites all green locally; lifecycle-test needs Node 20/22 — fails on Node 24 due to a lodash `dupMap` incompatibility, pre-existing on master, CI runs green)
- `tsc --noEmit` clean, eslint clean, prettier clean

## Status — 2026-08-05

| Step | State |
|---|---|
| Unit tests | touch/drop/responsive suites pass |
| tsc / eslint / prettier | Clean |

## Summary by CodeRabbit

* **Improvements**
  * Improved external item drops for more consistent placement and completion across mouse, HTML5, and touch interactions.
  * Enhanced touch drop handling so items are committed at their latest position after movement.
  * Reduced unnecessary layout updates while preserving existing grid behavior.
  * Simplified drop-position handling for more reliable placement.

### DIFF
--- a/src/react/components/GridLayout.tsx
+++ b/src/react/components/GridLayout.tsx
@@ -834,6 +834,16 @@ export function GridLayout(props: GridLayoutProps): ReactElement {
           droppingPositionRef.current.top !== newDroppingPosition.top;
         if (shouldUpdate) {
           setDroppingPosition(newDroppingPosition);
+          // Keep the placeholder's layout entry in sync so commitDrop/onDrop
+          // reads the latest cell, not the initial one (#2284).
+          setLayout(currentLayout =>
+            currentLayout.map(item =>
+              item.i === itemToPlace.i &&
+              (item.x !== calculatedXY.x || item.y !== calculatedXY.y)
+                ? { ...item, x: calculatedXY.x, y: calculatedXY.y }
+                : item
+            )
+          );
         }
       }
     },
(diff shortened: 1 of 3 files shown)


STRML:
@coderabbitai review

Issue #2210, opened 2025-12-23
v2.1.1 still has the 'Warning: Maximum update depth exceeded' when dragging from outside.


fpsqdb:
### Describe the bug

The warning is triggered when dragging an item into the grid and then moving it outside without releasing the mouse button.

### Your Example Website or App

https://codesandbox.io/p/devbox/react-grid-layout-forked-442dz3?file=%2Fsrc%2FShowcaseLayout.jsx%3A113%2C15&workspaceId=ws_Q9mnfkAHJiJStL3MgbyNFY

### Steps to Reproduce the Bug or Issue

drag item

### Expected behavior

no error

### react-grid-layout library version

2.1.1

### Operating System Version

windows 11

### Browser

Chrome

### Screenshots or Videos

https://github.com/user-attachments/assets/1743c621-942e-46d1-ab4d-b338d734977d

fpsqdb:
@STRML I have tested version 2.2.0, but the issue persists. Here is the reproduction link:

https://codesandbox.io/p/devbox/react-grid-layout-forked-vgyhsl?workspaceId=ws_Q9mnfkAHJiJStL3MgbyNFY

https://github.com/user-attachments/assets/91cf30d1-3f46-4f27-91de-42d100d05b0d

fpsqdb:
@STRML There are other bugs in v2.2.0: the position of the item shifts/offsets when dragging starts.

https://github.com/user-attachments/assets/a9dc5c10-7295-4b72-a273-1d1af0bf46a5

fpsqdb:
@STRML This bug is still there in v2.2.1.

https://codesandbox.io/p/devbox/react-grid-layout-forked-6fg7th?workspaceId=ws_Q9mnfkAHJiJStL3MgbyNFY

https://github.com/user-attachments/assets/479bbba4-6631-495a-beda-dee2f08c1c31

STRML:
Can confirm this is finally fixed. https://codesandbox.io/p/devbox/react-grid-layout-forked-6fg7th

PR #2287, opened 2026-08-06
fix: do not fire onLayoutChange mid-external-drag (#2219)

### LINKED ISSUE #2219: Meta: onLayoutChange fires prematurely during external drag operations
## Summary

This is a tracking issue for the behavior where `onLayoutChange` fires during external drag operations (drag-from-outside) before an item is actually dropped. This causes problems when users save layout state in `onLayoutChange`, as the layout gets saved during drag enter/leave events rather than only on actual drops.

## Related Issues

- **#1862** - `onLayoutChange` called when something is dragged in/out layout (but not yet dropped)
  - Author: @avasuro
  - Core issue describing the problem
  
- **#1893** - Items get misplaced when an element is not dropped in the layout
  - Author: @rishavbharti
  - Symptom: items move positions when dragging in/out without dropping
  - Related to `allowOverlap` state being lost (partially fixed in #2216)

## Current Workaround

Use `onDrop`, `onDragStop`, and `onResizeStop` callbacks instead of `onLayoutChange` for persisting layout state:

```jsx
// Instead of:
// onLayoutChange={handleLayoutChange}

// Use:
onDrop={handleLayoutChange}
onDragStop={handleLayoutChange}
onResizeStop={handleLayoutChange}
```

## Proposed Solution

Consider one of:
1. Add an option like `onLayoutChangeOnDropOnly` or a flag to control when `onLayoutChange` fires
2. Change the default behavior to only fire `onLayoutChange` after drop completes
3. Add a separate `onLayoutPreview` callback for during-drag state changes

## Affected Users

If you're affected by this issue, please 👍 this issue or comment with your use case.

cc: @avasuro @rishavbharti @mriabokliach

STRML:
## Summary

Fix #2219: `onLayoutChange` no longer fires during external drags (HTML5 or touch). The dropping placeholder is transient gesture state — it enters and leaves the layout during hover — but the layout-change effect fired `onLayoutChange` on every drag-enter and drag-leave, before any item was dropped. Consumers persisting layout state in `onLayoutChange` saved transient placeholder state.

## Root cause

`handleDragOver` / the touch adapter add the dropping placeholder to `layout` during hover but never set `activeDrag`. The layout-change effect (`GridLayout.tsx`) fires whenever `layout` changes and `!activeDrag`. So every placeholder add/remove triggered `onLayoutChange`.

## Fix

Suppress the callback while the placeholder is active — symmetric with how `activeDrag` already suppresses it during internal drags. The placeholder is already filtered from the payload (#2210), so firing for it reported a layout whose only change was withheld. The committed drop removes the placeholder, so the post-drop layout change fires normally. No new API, no breaking change.

## Tests

- **touch-drop-test.tsx**: a touch gesture (touchstart → touchmove → touchend off-grid) no longer fires `onLayoutChange` — proven to fail without the fix.
- **lifecycle-test.js**: drag-over renders the placeholder but `onLayoutChange` only fires after a real drop.

## Status — 2026-08-06

| Step | State |
|---|---|
| Unit tests | 529 pass + new regression, tsc clean |
| Formatting/lint | Prettier + ESLint clean |

Note: `lifecycle-test.js` can't load on Node 24 locally (pre-existing jest-haste-map `ModuleMap._assertNoDuplicates` when importing the built dist) — CI runs Node 20/22 and exercises it. The touch-drop suite exercises the same code path locally.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented layout change notifications during active drag, resize, and external drop-preview interactions.
  * Layout updates are reported after drop operations complete, excluding temporary placeholder changes.
  * Improved touch-drop behavior by preventing callbacks when temporary placeholders are created or removed.

### DIFF
--- a/src/react/components/GridLayout.tsx
+++ b/src/react/components/GridLayout.tsx
@@ -489,7 +489,20 @@ export function GridLayout(props: GridLayoutProps): ReactElement {
 
   // Layout change callback
   useEffect(() => {
-    if (!activeDrag && !deepEqual(layout, prevLayoutRef.current)) {
+    // Suppress while a gesture is mid-flight: activeDrag covers internal
+    // drags/resizes, droppingDOMNode covers external drags (the placeholder is
+    // added/removed during hover but no real item moves — placeDroppingItem
+    // appends without compacting). Firing here would report a layout whose only
+    // change is the transient placeholder, which is filtered from the payload
+    // anyway (#2210, #2219). The committed drop removes the placeholder, so the
+    // post-drop layout change fires normally. Skipping the prevLayoutRef sync
+    // here is intentional — the suppressed states must not desync the
+    // comparison, or the real change would be compared against the placeholder.
+    if (
+      !activeDrag &&
+      !droppingDOMNode &&
+      !deepEqual(layout, prevLayoutRef.current)
+    ) {
       prevLayoutRef.current = layout;
       // Filter out dropping placeholder - it's transient internal state only (#2210)
       // The dropping item should not be exposed to users until the actual drop happens.
@@ -498,7 +511,7 @@ export function GridLayout(props: GridLayoutProps): ReactElement {
       const publicLayout = layout.filter(l => l.i !== droppingItem.i);
       onLayoutChange(publicLayout);
     }
-  }, [layout, activeDrag, onLayoutChange, droppingItem.i]);
+  }, [layout, activeDrag, droppingDOMNode, onLayoutChange, droppingItem.i]);
 
   // ============================================================================
   // Container Height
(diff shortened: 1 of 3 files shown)


STRML:
> **Unresolved: use Testing Library events in the droppable test**

`TestUtils.Simulate` is used in 9 places across `lifecycle-test.js` (lines 577-844), not just the test this PR touches. Migrating all of them to `fireEvent` is a separate cleanup beyond this fix's scope; my change reuses the existing `dragDroppableTo` helper because it's the established pattern in that file. The `TestUtils` import can't be removed without touching the other 8 call sites.

The `elementFromPoint` restoration finding is fixed in d4c5507.

